PR #5, opened 2021-07-01
chore: Config changes to get static build working

### LINKED ISSUE #1: Set up NextJS in portal client repo and configure to build to static site
- Bootstrap a NextJS app
- Document the process to build NextJS to a static site

suzubara:
## Description 

Continue setting up NextJS so we can build to a [static site](https://nextjs.org/docs/advanced-features/static-html-export).

This adds the `yarn build:static` command, which builds and exports static assets into the `/out` directory.

Also a few other tweaks:
*  Enable absolute import paths (ie, `import Logo from 'components/Logo'` instead of `import Logo from '../components/Logo'`)
* Move test files for `/pages` into a top level `./__tests__/pages` directory because NextJS [does not like non-page files in its pages directory](https://nextjs.org/docs/messages/prerender-error) ☹️
* Adds file mock for assets in Jest ([ref](https://jestjs.io/docs/webpack#handling-static-assets))
* Added bundle analyzer to the webpack config which we can use to track JS bundle size & composition (use `yarn build:analyze` to generate)

Fixes #1 

## Review Notes

You should be able to check this branch out and run `yarn build:static` to produce the static site in the `out` directory. You can then use a tool like [serve](https://github.com/vercel/serve) to serve the static site (ie, `serve -s out`).


components/Logo/Logo.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+import ussfLogo from 'public/img/ussf-logo.svg'
+
+const Logo = () => {
+  return <img src={ussfLogo} alt="USSF Logo" />
```
abbyoung, line 6: This url was rendering as [object object] for me -- turns out the svg imports as an object?! Adding `ussfLogo.src` should fix it.

suzubara: omg I was wrestling with this at the end of the day yesterday 😭 THANK YOU

.eslintrc.json
```diff
@@ -38,16 +38,20 @@
     "import/no-extraneous-dependencies": "error",
     "import/order": "warn",
     "react/button-has-type": "error",
-    "@typescript-eslint/explicit-module-boundary-types": "off"
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@next/next/no-img-element": "off"
```
suzubara, line 42: I'm disabling this rule after wrestling with the `next/image` component and because it is [not supported in static exports](https://nextjs.org/docs/messages/export-image-api) without implementing a custom loader. We can still _use_ `next/image`, I just don't want ESLint telling us we should _always_ use it when it doesn't work for our current use case. my thought is let's come back to this once we know more about our deployment / CDN situation (there is an [akamai loader](https://nextjs.org/docs/basic-features/image-optimization#loader) we might be able to use)

__tests__/pages/index.test.tsx
```diff
@@ -3,7 +3,7 @@
  */
```
suzubara, line 3: I had to move this file because NextJS wants _everything_ inside `/pages` to actually be a page

pages/_app.tsx
```diff
@@ -1,5 +1,3 @@
-import '@trussworks/react-uswds/lib/uswds.css'
-import '@trussworks/react-uswds/lib/index.css'
```
suzubara, line 2: removed these to minimize the bundle for now, we can re-add them when we set up USWDS

pages/index.tsx
```diff
@@ -1,5 +1,11 @@
 import Head from 'next/head'
 
+import Logo from 'components/Logo/Logo'
+
+export const config = {
+  unstable_runtimeJS: false,
```
suzubara, line 6: this is an [experimental flag](https://github.com/vercel/next.js/pull/11949) I discovered while looking into how to minimize Next's export bundle. We can use this to prevent NextJS loading scripts to "rehydrate" the client for pages with truly no JS. There's more discussion about this feature [here](https://github.com/vercel/next.js/issues/5054). We will probably end up removing this once we start adding actual client-side JS but for now it keeps the exported site as small as possible!

components/Logo/Logo.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const Logo = () => {
+  return <img src="/img/ussf-logo.svg" alt="USSF Logo" />
```
abbyoung, line 4: I'm still not able to see this image -- am I missing something? ☹️

abbyoung: JUST KIDDING it's there, it's beautiful

jest.setup.js
```diff
@@ -1 +1,16 @@
+/* eslint-disable react/react-in-jsx-scope */
+/* eslint-disable no-import-assign */
+/* eslint-disable react/display-name */
+
 import '@testing-library/jest-dom'
+import * as NextImage from 'next/image'
+
+// There are some open issues with NextImage in Jest:
+// https://github.com/vercel/next.js/issues/26749
+// https://github.com/vercel/next.js/issues/26606
```
abbyoung, line 10: wow thank you for finding/including these issues so we can track!

abbyoung (approved):
lgtm 🎉 was able to build static site and serve

esacteksab:
> lgtm 🎉 was able to build static site and serve

Same!

esacteksab (approved):
It's _tiny_!

```shell
➜  ussf-portal-client git:(1-build-static-site) ✗ tar -czf nextjs-test.tgz out 
➜  ussf-portal-client git:(1-build-static-site) ✗ ls -lah nextjs-test.tgz 
-rw-r--r--  1 tempuser  staff    99K Jul  2 13:35 nextjs-test.tgz
```

Issue #22, opened 2021-07-02
Set up Cypress CI in Github Actions


suzubara:
https://docs.cypress.io/guides/continuous-integration/github-actions#Cypress-GitHub-Action

suzubara:
@esacteksab The Cypress dev team has an official Github action but it is not "verified" by Github: https://github.com/marketplace/actions/cypress-io

Are you ok with us adding it to our allowed actions?

esacteksab:
Looks like they're trying to get verified https://github.com/cypress-io/github-action/issues/305#issuecomment-832021887

Nothing sticks out. Let's pin to a specific release and go for it. 

Thanks!

PR #25, opened 2021-07-06
test: Add Cypress with basic home page test

### LINKED ISSUE #21: Set up Cypress testing
Set up [Cypress](https://docs.cypress.io/) in this project so we can use it for integration & E2E tests of the static site

suzubara:
## Description 

This PR adds [Cypress](https://docs.cypress.io/) to the project, with one basic test on the home page. Setting up Cypress to run as part of CI is covered in https://github.com/USSF-ORBIT/ussf-portal-client/issues/22 and it may be that we decide to move Cypress into its own repo, but for now this will let us start writing integration tests early and easily, which will be an important part of the MVP rebuild process.

Fixes #21 

## Review Notes

* Check out this branch and run `yarn install` to install new dependencies
* The dev server has to be running: `yarn dev`
* Start the Cypress test runner UI: `yarn cypress:dev`
* Run the homepage integration test - it should pass


abbyoung (approved):
verified setup, home integration tests pass 🎉

esacteksab (approved):

PR #27, opened 2021-07-07
feat: Migrate `head` content, meta data from MVP site

### LINKED ISSUE #4: Migrate meta data and static assets from MVP
Make sure the `<head>` meta tags and associated assets (icons, etc.) mirror those from the MVP site: https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/_includes/head.html

suzubara:
## Description 

Adds all of the `head` elements, meta data, icons, etc. that are currently part of the MVP site (ref: https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/_includes/head.html)

* Common & default `<head>` elements are defined in `pages/_app.tsx`. These can all be overridden by passing elements into `<Head>` on a page component (see the Test page example).

Fixes #4 

## Review Notes

* This introduces our first environment variable (see [NextJS docs](https://nextjs.org/docs/basic-features/environment-variables), it didn't seem to like looking for an `.envrc` file). If there is a better way to store the canonical domain per-environment I'm open to feedback. If not, I'll amend this PR to include docs about copying `.env.local.template` as part of dev setup.
* To test, visit the site locally and verify that the expected title & description for the home page and test page are visible in the browser window/source code. There's also a Cypress test to look for these.
* You should also verify none of the public assets result in a 404 error (such as the device-specific icons).
  * The favicon should show up in your browser window.
  * You can test the Safari tab icon by opening the site in Safari and opening another tab (see screenshot below).
  * You can also test iOS/android assets on a physical device by navigating to your computer's local IP at port 3000 when running the dev server. The icons should be used when saving the site to your home screen (see screenshots below).

image

IMG_6411
IMG_6412


suzubara:
note to self: update documentation about `.env.local` before merging

cypress/integration/home.spec.js
```diff
@@ -1,6 +1,35 @@
...
     cy.contains('USSF Portal')
   })
+
+  it('contains the expected meta data', () => {
+    cy.document()
+    cy.get('head title').should('contain', 'Space Force Portal')
+    cy.get('head link[rel="canonical"]').should(
+      'have.attr',
+      'href',
+      'http://localhost:3000/'
```
abbyoung, line 16: flagging as question for eng sync today 🏳️

public/site.webmanifest
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "",
```
abbyoung, line 2: Should we include the name of the site here, or is this primarily for icons?

pages/index.tsx
```diff
@@ -1,24 +1,25 @@
-import Head from 'next/head'
+// https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md#case-i-use-nextjs-and-im-getting-this-error-inside-of-links
```
abbyoung, line 1: I really appreciate including this documentation along with the eslint-disable! What's the best way to track moving forward, to make sure that if/when the API is updated, we also update this code?

.env.local.template
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_SITE_URL='http://localhost:3000'
```
abbyoung, line 1: Do we need to have this env variable to the Netlify deployment as well?

abbyoung: (obvs different value but you know what I mean)

public/site.webmanifest
abbyoung: fwiw it looks like whatever is set as the short_name is used when saving to an iPhone

.env.local.template
abbyoung: brain mush -- GOT IT.

pages/index.tsx
abbyoung: discussed in eng sync 🎉

public/site.webmanifest
abbyoung: discussed in eng sync 🎉

abbyoung (dismissed):
lgtm! thanks for talking through everything with me!

pages/_app.tsx
```diff
@@ -1,8 +1,56 @@
 import 'styles/index.scss'
 import type { AppProps } from 'next/app'
+import Head from 'next/head'
```
esacteksab, line 3: You import `Head` from `'next/head'` but where is `'next/head'`?

suzubara: anything imported from `next/*` is from the nextJS package

esacteksab (approved):

esacteksab (approved):

abbyoung (approved):

Issue #29, opened 2021-07-09
Create a NavLink component


suzubara:
Create a NavLink component that incorporates `useRouter` to the existing Link component, and adds `activeClassName` and `exact` props in order to style navigation links that are currently being viewed.

for reference: https://stackoverflow.com/questions/60324081/adding-active-class-for-nav-link-in-next-js

Issue #28, opened 2021-07-09
Create a Link wrapper component


suzubara:
Create a wrapper component for `next/link` that can be added to eslint config to get around the anchor a11y issue (see the 3rd option described here, with example code linked: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md#case-i-use-nextjs-and-im-getting-this-error-inside-of-links)

PR #30, opened 2021-07-09
dev: MVP Header component

### LINKED ISSUE #8: Create shared header / navigation component
Create a shared component that renders the navigation, mobile navigation, and optional page navigation & mirrors the MVP header component: https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/_includes/header.html

suzubara:
## Description 

This PR adds a Header component that implements the [MVP header](https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/_includes/header.html) using React and ReactUSWDS. It also copies over relevant global CSS from the MVP codebase in order to match the styling as closely as possible.

I created a couple of followup issues in order to clean up some of the Link code:
- #28 
- #29 

Fixes #8 

## Review Notes

You can view the component in Storybook:
- `yarn storybook`

image

image

image


suzubara:
@abbyoung Do you think we should namespace the MVP components (ie, `/components/mvp`) to more easily distinguish between them once we start building new/beta components?

abbyoung:
> @abbyoung Do you think we should namespace the MVP components (ie, `/components/mvp/header`) to more easily distinguish between them once we start building new/beta components?

Yeah I think that's a great idea, especially since we have no clue what the new/beta components will be! I like how you separated out the mvp css too.

abbyoung:
It looks like existing version is using Public Sans Web as its font-mono default and we've got Source Sans Pro in Storybook.

Also is it something weird with Storybook that they both are set to font size 12px, but in Storybook it looks significantly smaller? I don't think it's just the font discrepancy, but maybe.

components/MVP/Header/Header.tsx
```diff
@@ -0,0 +1,146 @@
...
+    { path: '/#manage-your-life', label: 'Manage your life' },
+    { path: '/#work-tools', label: 'Work tools' },
+  ].map((i) => (
+    <Link href={i.path} key={i.path}>
+      {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+      <a className="usa-nav__link">{i.label}</a>
+    </Link>
+  ))
+
+  const servicePortalMenu = (
```
abbyoung, line 76: This is existing behavior in the deployed mvp, but the dropdown isn't trapping focus. Would we like to do that moving forward when building any dropdowns? Curious what the best practice is.

abbyoung (approved):
lgtm 🎸

components/MVP/Header/Header.tsx
suzubara: we talked about this, the same issue occurs in USWDS dropdowns. For future development we may look into using a focus trap for elements like this.

suzubara:
@abbyoung I dug a little and found some SCSS variables that had different values in the MVP! I think this will fix the mismatched fonts. I also figured out how to get Storybook to share the same settings file so we don't have to duplicate them 🎉

styles/storybook.scss
```diff
@@ -1,21 +1,15 @@
...
 
-$theme-show-notifications: false;
-/* GENERAL */
-$theme-image-path: '~uswds/dist/img';
+// ReactUSWDS CSS modules
+@import '@trussworks/react-uswds/lib/index.css';
 
-/* TYPOGRAPHY */
-$theme-font-path: '~uswds/dist/fonts';
+$asset-path: '~uswds/dist';
```
abbyoung, line 6: so satisfying!!!

styles/_settings.scss
```diff
@@ -0,0 +1,24 @@
+$asset-path: '' !default;
```
abbyoung, line 1: 🎉

abbyoung (approved):

PR #37, opened 2021-07-16
feat: Default page layout

### LINKED ISSUE #11: Create shared page layout
Create a shared page layout component that mirrors the one in the MVP: https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/_layouts/default.html

Except for the following changes:
- It should render the ReactUSWDS GovBanner component instead of implementing one from scratch
- It should not include page-specific JS (this can be included in page components)
- It should not include USWDS JS

suzubara:
## Description 

This PR creates a `DefaultLayout` component and implements it in the NextJS app so it will wrap all page content automatically ([this can be bypassed on a per-page basis as needed](https://nextjs.org/docs/basic-features/layouts#per-page-layouts)).

I also ran into some issues resolving CSS `url()` asset paths, so there are also some changes there. The biggest one is that I added a shell script to copy over the two asset directories from USWDS (containing fonts & images) as a [postinstall](https://docs.npmjs.com/cli/v7/using-npm/scripts) script, which will run automatically on every `yarn install`. It will copy those folders into the `/public` directory, which is what NextJS uses to serve static assets. Ideally this would be handled automatically by webpack, but I reached my limit of how long I wanted to try and get that working to no avail. On the flipside, this means Storybook can use the same asset paths as NextJS, so I was able to remove the Storybook-specific SCSS variables.

Fixes #11 

## Review Notes

You should be able to run both the NextJS app & Storybook sites without error:
```
yarn dev
yarn storybook
```

The home page (and test page) both have no real content, but they should both now render within the shared default layout, which should match that on the existing MVP portal. The shared elements to look for are:
- header & navigation
- USA gov banner
- footer
- identifier

image

image

image


abbyoung (approved):
🎉

PR #39, opened 2021-07-19
refactor: Move app code into src directory

### LINKED ISSUE #36: Move application code into `/src`
By default all of our application code is located in the root directory, but NextJS also [supports the common pattern of locating it in a `/src` directory](https://nextjs.org/docs/advanced-features/src-directory). As we start building out this codebase and add on other tools like Storybook & Cypress, I think it would benefit our organization to move app code into `/src`. The following directories should be moved:
- `__mocks__`
- `__tests__`
- `/components`
- `/layout`
- `/pages`
- `/styles`

abbyoung:
## Description 

*  Moves app code into `src/` directory
* Updates paths as needed
* Updates Storybook `preview.js` to resolve path correctly

Fixes #36 

## Review Notes
Make sure you can still run `yarn dev` and `yarn storybook` without any errors.


src/__tests__/pages/index.test.tsx
```diff
@@ -1,6 +1,7 @@
 /**
  * @jest-environment jsdom
  */
+import React from 'react'
```
abbyoung, line 4: I had to add this import in a number of places to appease the linting gods. Is there a better way to handle?

suzubara: Yes! You can add (update) this path in `eslintrc.json` in the `overrides` array

suzubara (dismissed):

src/__tests__/pages/index.test.tsx
abbyoung: Updated -- thank you so much!

suzubara (approved):

PR #40, opened 2021-07-19
refactor: Create LinkTo utility component

### LINKED ISSUE #28: Create a Link wrapper component

abbyoung:
## Description 

Adds a utility component to wrap the Next.js `<Link/>` component and pass the props as necessary to `<Link/>` and `<a>`. This is a workaround for a known issue in Next.js, and allows us to write code without inline eslint disable rules.

*  Uses [utility component](https://gist.github.com/zackdotcomputer/d7af9901e7db87364aad7fbfadb5c99b) shared in the jsx-eslint docs
* Adds tests for component
* Replaces applicable instances of `<Link/>` with `<LinkTo/>`

Fixes #28 

## Review Notes

* Currently being used for external links in Header and Footer components. Please test/check and make sure everything is as expected.

* The utility component written as-is works great and meets our needs. If there any other mods we want to make @suzubara, please let me know!

* Do I need to create a directory for utility components? We'll at least have one another, maybe more. Wasn't sure if it was okay to keep all together or if it makes sense to separate.


suzubara (commented):
this looks great! I like the idea of organizing utility components, maybe in `components/util`? Also I think there are still a few instances of `<Link />` in the header component that can be replaced with `LinkTo`

suzubara (approved):

PR #41, opened 2021-07-21
ci: Set up Cypress E2E tests in Github Actions

### LINKED ISSUE #22: Set up Cypress CI in Github Actions

suzubara:
## Description 

This PR uses the [Cypress.io Github Action](https://github.com/marketplace/actions/cypress-io) to run Cypress tests against the static site on PRs & pushes to the `main` branch. Artifacts will be stored in Github for debugging failures.

Fixes #22 

## Review Notes

Verify the [Cypress check](https://github.com/USSF-ORBIT/ussf-portal-client/pull/41/checks) runs on this PR and passes.

You can also simulate the Cypress workflow locally by doing the following:
```
yarn install
NEXT_PUBLIC_SITE_URL=http://localhost:5000 yarn build:static
yarn start:static
CYPRESS_BASE_URL=http://localhost:5000 yarn cypress run
```


cypress.json
```diff
@@ -1 +1,3 @@
-{}
+{
+  "baseUrl": "http://localhost:3000"
```
esacteksab, line 2: You have `localhost:3000` here but above, you set `CYPRESS_BASE_URL` to `localhost:5000`, are these not the same?

suzubara: They are not the same - I attempted to add this detail to the README: https://github.com/USSF-ORBIT/ussf-portal-client/blob/125f9fcbf9db298e6eeca69f681a8202e1ab3af2/README.md#static-site-vs-nextjs-server

tldr is `localhost:5000` is where the static site is served, and since that most closely represents what will be in production when we deploy next, that's what I have Cypress testing against. `localhost:3000` is the NextJS server & what is running during local development, which is why it's set as the default here.

abbyoung (approved):
🎉

PR #42, opened 2021-07-21
refactor: NavLink component

### LINKED ISSUE #29: Create a NavLink component

suzubara:
## Description 

Adds a `NavLink` component, which builds on the `LinkTo` component and has a dependency on `next/router` in order to apply an "active" CSS class to links that match the current path.

I also replaced any outstanding uses of `next/link`.

Fixes #29 

## Review Notes

- In the app (`yarn dev`) verify each of the Header links have the "active" style when you click on them (even if they 404)
- In Storybook (`yarn storybook`) verify the Home & News links in the Header component have the "active" style on their respective stories


src/components/MVP/Header/Header.test.tsx
```diff
@@ -112,7 +112,7 @@ describe('Header component', () => {
 
   describe('on the news page', () => {
     beforeEach(() => {
-      mockedUseRouter.mockReturnValueOnce({
+      mockedUseRouter.mockReturnValue({
```
suzubara, line 115: This has to be returned multiple times since `useRouter` now is called on both the Header & NavLink components

src/components/MVP/Header/Header.tsx
```diff
@@ -40,15 +40,13 @@ const Header = () => {
...
-        <span>{i.label}</span>
-      </a>
-    </Link>
+    <NavLink
+      href={i.path}
+      key={i.path}
+      className={classnames('usa-nav__link', i.className)}
+      exact={i.path === '/'}>
+      <span>{i.label}</span>
+    </NavLink>
```
abbyoung, line 49: It looks like the existing mvp doesn't assign an active class to the Home link when clicked. I'm fine with keeping it for consistency, although it does disappear once one of the 'On this page' links are clicked.

abbyoung (approved):
Tested and looks good/works great! Left one comment but not a blocker.

src/components/MVP/Header/Header.tsx
suzubara: @abbyoung Ohh good catch, I didn't even notice! @noahfirth how do you feel about this?

noahfirth: Let's not make an exception, I kind of like how it looks with the active state marker next to it on the home page.

PR #45, opened 2021-07-23
feat: Home Page

### LINKED ISSUE #16: Build the Home page
Page URL: `/` (the index)

Includes content:
- COVID 19 alert (https://github.com/USSF-ORBIT/ussf-portal-client/issues/10)
- PT test and CSO townhall announcement cards
- Static content:
  - Quick Links (myPay, DTS) - sidebar
  - Manage your life content
  - Work Tools content
  - Learn and Grow content
- Links to other service portals

Mirrors https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/pages/about.md

abbyoung:
## Description 
 
* Adds rebuild of MVP Home page. Content and styling should match the current deployed Home page
* Adds unit and integration tests
* Adds workaround for font awesome icon rendering

Fixes #16

## Review Notes

Run the app locally with `yarn dev` and verify the Home page looks as expected.


src/components/MVP/AnnouncementCard/AnnouncementCard.module.scss
```diff
@@ -25,12 +27,20 @@
...
+    .usa-card-group + h2 {
+      margin-top: units(6) !important;
+    }
     .usa-card__container {
       &:hover {
         background-blend-mode: soft-light;
         border-color: color('primary-light');
       }
     }
+    .usa-tag {
```
abbyoung, line 39: Not sure why this is reverting back, I must have some funky commits happening. Will get it fixed up.

src/pages/index.tsx
```diff
@@ -1,14 +1,256 @@
...
+                </ul>
+              </div>
+            </div>
+          </div>
+        </div>
+      </section>
+
+      <section
+        className={`usa-section padding-top-6 padding-bottom-0 bg-base-lightestx ${styles.home}`}>
+        <div className="grid-container usa-prose">
```
abbyoung, line 124: @suzubara I noticed the Grid components in your About page PR and probably should've used it in mine... 🙈 Still getting used to the React USWDS components that are available.

suzubara: no problem, I like to use them where I can but they also don't fit every scenario so I wouldn't worry too much about it!

src/components/MVP/CovidSiteAlert/CovidSiteAlert.tsx
```diff
@@ -7,6 +7,9 @@ import styles from './CovidSiteAlert.module.scss'
 const CovidSiteAlert = () => {
   return (
     <SiteAlert
+      //#TODO: Best way to check that a component rendered
+      // on a page re: testing? This seems easiest.
```
suzubara, line 11: This is a good reference for best ways to select components: https://testing-library.com/docs/queries/about#priority

_also_ if you use `getByRole()` in a test but there's no result, it will helpfully return the list of roles that are present on the page! that is usually how I will figure out if there's a role I can use if I'm not sure. so I tried that in the test and found this result:
```
Here are the accessible roles:

      region:

      Name "Site alert":
      <section
        aria-label="Site alert"
        class="usa-site-alert usa-site-alert--info usa-site-alert--no-heading usa-site-alert--no-icon covidSiteAlert"
        data-testid="covid-alert"
      />
```

so in this case I think it can use `screen.getByRole('region', { name: 'Site alert' })` 🎉

src/pages/_app.tsx
```diff
@@ -49,6 +49,8 @@ const USSFPortalApp = ({ Component, pageProps }: AppProps) => {
 
         <meta name="msapplication-TileColor" content="#da532c" />
         <meta name="theme-color" content="#ffffff" />
+        {/* Fix Next.js's rendering of font awesome css*/}
+        <style>{dom.css()}</style>
```
suzubara, line 53: wat

abbyoung: LOL. i could not get the icon to render the correct size without this workaround, but considering you just merged in a PR with an icon that looked just fine, I need to revisit.

suzubara: I did have to add some CSS to that icon to get it to look okay though, not sure which is better

abbyoung: for ref: https://github.com/FortAwesome/react-fontawesome/issues/284

suzubara (approved):
This looks awesome!! 🎉

suzubara (approved):

PR #56, opened 2021-07-26
chore: Disable NextJS telemetry using env flag

### LINKED ISSUE #54: Turn off NextJS telemetry (https://nextjs.org/telemetry)
NextJS collects anonymous telemetry data by default. We would like to disable this.

suzubara:
## Description 

[NextJS collects telemetry data by default](https://nextjs.org/telemetry) on certain commands. This can be disabled using an environment flag, or by running `npx next telemetry disable` on a per-machine basis. This is not ideal, but NextJS does not currently support opting out using their config file, so it has to be disabled for each environment.

Fixes #54 

## Review Notes

This NextJS functionality is pretty opaque, but the best way I could find to verify the flag is working as intended is to view the logs for the Cypress tests in Github, when the `next build` command is run. Since it's always a fresh environment, the telemetry message always appears ([see this example](https://github.com/USSF-ORBIT/ussf-portal-client/runs/3163402793?check_suite_focus=true#step:4:62)):
> Attention: Next.js now collects completely anonymous telemetry regarding usage.
This information is used to shape Next.js' roadmap and prioritize features.
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://nextjs.org/telemetry

If you check the same place in the logs for Cypress on this branch, there should be no such message, which based on the NextJS source code indicates the telemetry is disabled.


abbyoung:
Should this also work for my local machine, or do I still need to disable it? Looks like if I run `npx next telemetry status` locally it's enabled, so I'm assuming I need to disable.

The Cypress logs check out!

suzubara:
> Should this also work for my local machine, or do I still need to disable it? Looks like if I run `npx next telemetry status` locally it's enabled, so I'm assuming I need to disable.
> 
> The Cypress logs check out!

I had the same experience, where `npx next telemetry status` returns enabled even if my environment variable is set, so I don't necessarily trust it 🤔  I tried looking through NextJS issues & their source code and it seems people have various issues with the way this works now. It's really annoying that it's not just something we can set in the next config file. One alternative is to run `npx next telemetry disable` as a package.json lifecycle script but I'm not sure I like that any more 😕

abbyoung (approved):
lgtm 👍 we can always monitor this config/feature from nextjs and see if there's a better way to handle in the future 😩

PR #62, opened 2021-07-29
build: Set up docker

### LINKED ISSUE #63: Set up Docker for running the app
Set up Docker so it can be used for:
- Running the app in dev mode with hot reloading
- Building the app static assets
- Running the app in production mode

suzubara:
## Description 

This PR sets up Docker to run the portal application in both development & production modes. It also switches env vars from `.env` to `.envrc` (and updates relevant documentation) and updates to the latest node 14 version.

Right now there is not a compelling reason to run the app in docker during development, but once we start adding other services (like a db) it will probably come in handy. It is also useful as a mirror of deployed environments.

Fixes #63 

## Review Notes

- Verify you can run the app in Docker in dev mode according to the instructions
  - In this mode, you should be able to make a change in the application code and it should reload with the changes
  - When you open the site and inspect React components in the dev tools, they should not be minified
- Verify you can run the app in Docker in production mode according to the instructions
  - You'll have to set the env vars: `NODE_ENV=production, TARGET_ENV=production`
  - When you open the site and inspect React components in the dev tools, they should be minified


abbyoung (approved):
it works! great documentation, had no problems setting up Docker and running the app. 🚀

esacteksab (approved):

PR #64, opened 2021-08-04
[Experimental][WIP] Add embedded Keystone

suzubara:
## Description 

This is an experiment, DO NOT MERGE!

* `yarn install && yarn dev` will start the client at localhost:3000 and the Keystone UI at localhost:8000
* Also added the graphQL explorer at http://localhost:3000/api/graphql
* All Keystone config is done in `keystone.ts`. The GraphQL & Prisma schemas are generated by `keystone-next postinstall` (but should still be checked into the code)

More info on embedded mode: https://keystonejs.com/docs/walkthroughs/embedded-mode-with-sqlite-nextjs

This is important to note:
> * Because Admin UI access is restricted to a local development runtime, this mode is (currently) not suitable when you need more than one editor.
> * SQLite doesn’t support scalability, concurrency, and centralisation as well as PostgreSQL. If your project has scaling needs, you’re probably better off with standalone mode and external PostgreSQL hosting.


suzubara:
Closing in favor of actual implementation: https://github.com/USSF-ORBIT/ussf-portal-client/pull/213

PR #67, opened 2021-08-05
ci: Fix Husky pre-commit hook

### LINKED ISSUE #66: Husky pre-commit hook is not running locally
Our pre-commit hook is not running anymore, investigate & fix

suzubara:
## Description 

We noticed our pre-commit hook (which formats & lints staged files) didn't seem to be running anymore. I ran the Husky installation steps again and it added some files that were not actually checked into the repo.

After investigating, I believe this was due to a change in Husky's file structure, where previously, installed hooks were located in `./husky/_/` which was gitignored (since they are always installed locally). In a more recent version, they changed to not being gitignored, and so I deleted `.husky/.gitignore` in [this commit](https://github.com/USSF-ORBIT/ussf-portal-client/commit/6af3c57e310d29a6b123628414fc190f88abc9d5) but _did not explicitly add the Husky files_, resulting in the directory being deleted altogether from the repo.

I also went ahead and ran the format & lint scripts on all files to catch any changes that might have fallen through the cracks.

Fixes #66 

## Review Notes

Check out this branch, make a commit that includes an eslint violation and verify that the pre-commit hook runs and prevents you from committing.


abbyoung (approved):
All fixed -- thank you!

PR #70, opened 2021-08-09
ci: Initial commit for Dependabot configuration

### LINKED ISSUE #69: Set up dependabot
Dependencies don't seem to be auto-updating in this repo. Investigate & resolve

esacteksab:
## Description 

Adds a `dependabot.yml`. Should update `docker` and `npm` both in the root directory and should update `npm` in the `cypress` directory. (I'm not sure if this path is correct, but I need the file to exist to test it.

Fixes #69

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ updates:
     time: "11:00"
   open-pull-requests-limit: 10
   labels:
-  - dependencies
+  - 'type: dependencies'
 - package-ecosystem: npm
   directory: "/cypress"
   schedule:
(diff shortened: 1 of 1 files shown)


suzubara (commented):
yay thanks! a couple small things:
- Can we use the `type: dependencies` label for these?
- ReactUSWDS requires we use a specific version of USWDS so we will want to ignore `uswds` updates (we can update it manually when ReactUSWDS has updates)

esacteksab:
> yay thanks! a couple small things:
> 
>     * Can we use the `type: dependencies` label for these?
> 
>     * ReactUSWDS requires we use a specific version of USWDS so we will want to ignore `uswds` updates (we can update it manually when ReactUSWDS has updates)

I think I've captured these.

suzubara (approved):
looks good! let's try it and see what happens..

Issue #60, opened 2021-07-28
Release v1 (MVP rebuild)


suzubara:
Once all MVP issues have been completed, we should release major version 1.0.0 using the script added in https://github.com/USSF-ORBIT/ussf-portal-client/issues/61

PR #71, opened 2021-08-10
ci: Set up standard-version release script

### LINKED ISSUE #61: Set up release script
Add a `yarn release` script (described in https://github.com/USSF-ORBIT/ussf-portal/blob/main/docs/adr/0003-versioning-releasing.md#example-workflow-1):

> The release script runs on the main branch (this could be automated in Github actions):
Update the changelog and bump the version number
Commit those changes to a new release branch called release-x.x.x
Open a PR of release-x.x.x into the main branch

We can use https://github.com/conventional-changelog/standard-version or similar tooling

suzubara:
## Description 

* Adds [standard-version](https://github.com/conventional-changelog/standard-version) to the project 
* Configures `yarn release` to skip pushing a tag and only include certain commit types in the changelog

Fixes #61 

## Review Notes

This PR does _not_ issue an actual release (that will come in https://github.com/USSF-ORBIT/ussf-portal-client/issues/60). You can check out this branch and run `yarn release --dry-run` to preview the script.

Question: I just used the default settings for which commit types to include in the changelog (just features & bug fixes), but should we include any of the others?


package.json
```diff
@@ -80,5 +86,60 @@
...
+      {
+        "type": "feat",
+        "section": "Features"
+      },
+      {
+        "type": "fix",
+        "section": "Bug Fixes"
+      },
+      {
+        "type": "perf",
```
esacteksab, line 124: Why would we want to hide this? This feels like something worth highlighting.

package.json
```diff
@@ -80,5 +86,60 @@
...
+        "hidden": true
+      },
+      {
+        "type": "refactor",
+        "hidden": true
+      },
+      {
+        "type": "revert",
```
esacteksab, line 132: I'd sort of want this, I could be like "oh, they reverted it in this release" rather than having to wade through all the commit messages/history. And a revert _does_ change the code.

package.json
```diff
@@ -80,5 +86,60 @@
...
+      {
+        "type": "ci",
+        "hidden": true
+      },
+      {
+        "type": "dev",
+        "hidden": true
+      },
+      {
+        "type": "docs",
```
esacteksab, line 112: I'm on the fence on this one. Can we order the tags or is the order based on the chronological ordering of the commit history? Cause I could see putting `docs` down at the bottom where folks may have lost interest by then or simply not care.

package.json
suzubara: 🤷‍♀️  I don't know, I'm fine with including it!

package.json
suzubara: I believe the order of sections follows the same order as they're configured here (which I made alphabetical). But I'm fine with doing this & including docs at the end!

suzubara:
@esacteksab Changes made! I don't have super strong feelings about what makes it into the changelog and what doesn't.. I think it ultimately comes down to who is the audience for the changelog, and I think for now that's mostly us.

esacteksab (approved):

abbyoung (approved):

PR #102, opened 2021-08-11
ci: ignore 15-alpine and 16-alpine

### LINKED ISSUE #69: Set up dependabot

esacteksab:
## Description 

We don't want Node 15 or Node 16. This will hopefully ignore the updates in the Docker base image.

Admittedly, I'm not _exactly_ sure how this syntax should work with `package-ecosystem: docker`. I thought what I had would ignore updates, but didn't. So I'm explicitly saying ignore `15-alpine` and `16-alpine` (I think/hope). Again, this needs to land in main to test. This _should_ close #100 as we don't want it.

Fixes #69


suzubara:
I wonder if the ignore dependency-name should be `node` or the complete `node:15-alpine`? the supported syntax here is unclear :( alternatively, I think node is the only dependency in our Dockerfile right now so maybe we can omit docker altogether?

esacteksab:
> I wonder if the ignore dependency-name should be `node` or the complete `node:15-alpine`? the supported syntax here is unclear :( alternatively, I think node is the only dependency in our Dockerfile right now so maybe we can omit docker altogether?

I wondered about this myself honestly. I couldn't remember if Docker did anything else besides the base image. For some reason, I was thinking maybe it did packages as well. Either way, I'm fine not managing it. BRB.

suzubara (approved):

PR #112, opened 2021-08-13
ci: Don't allow Dependabot to update between Major releases for NPM packages

### LINKED ISSUE #111: Do not allow Dependabot to upgrade Major versions
In an effort to automate all the things, but not create too big of a headache, we decided during our last eng sync to not allow Dependabot to update Major version releases. Only patch and minor. Make it so.

esacteksab:
## Description 

To hopefully ensure that Dependabot doesn't introduce unexpected breaking changes, we're not allowing it to update between major releases for NPM packages. This should do that. Though, like past dependabot PRs, this needs to land in `main` to test.

Fixes #111


abbyoung (approved):
wonderful! thank you!

PR #131, opened 2021-08-16
fix: Add en as default lang

### LINKED ISSUE #129: bug: add html lang to nextjs app

abbyoung:
## Description 
`cypress-audit` surfaced a bug where a default lang wasn't set. This PR adds a default language.

Fixes #129 

## Review Notes
Weirdness to discuss: The audit fails locally, but is passing in the gh workflow on another branch, without this fix. Need to figure out why.


abbyoung:
`Error: i18n support is not compatible with next export. See here for more info on deploying: https://nextjs.org/docs/deployment`

abbyoung:
Closing as a wontfix for mvp. Once we move from static -> server, we can pass in the i18n config to set the language. In the meantime, this can be our data point for testing that cypress a11y tests are working as expected.

PR #141, opened 2021-08-17
ci: Allow dependabot to manage all version updates, but auto-merge will only automate minor and patch versions.

### LINKED ISSUE #111: Do not allow Dependabot to upgrade Major versions

esacteksab:
## Description 

We decided to let Dependabot manage all version updates, even major. But we're configuring `auto-merge` to only manage minor and patch updates. Major updates will still be done manually.

Like the others, this needs to land in `main` to test.

Fixes #111

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,6 @@ updates:
   ignore:
       # We don't want to update uswds. This update will be handled manually as needed.
     - dependency-name: "uswds"
-      # We don't want packages to update between major versions. Patch and minor only. Hopefully ensuring no breaking changes.
-    - dependency-name: "*"
-      update-types: ["version-update:semver-major"]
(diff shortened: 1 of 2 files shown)


suzubara (approved):
I read the docs for https://github.com/ridedott/merge-me-action and this looks right to me! Could we call the workflow something a little more descriptive though, something like "Automerge Dependabot"?

esacteksab:
> I read the docs for https://github.com/ridedott/merge-me-action and this looks right to me! Could we call the workflow something a little more descriptive though, something like "Automerge Dependabot"?

Ask, and you shall receive.

suzubara (approved):
thank u! 🙏

PR #145, opened 2021-08-18
[WIP] [DO NOT MERGE] Experimental POC using rewrites for beta routes

suzubara:
## Description 

This is an experimental proof of concept for using [NextJS rewrites](https://nextjs.org/docs/api-reference/next.config.js/rewrites) to handle beta routes.

This example only uses client-side cookies (sets & deletes `optInBeta` in the button click handlers) -- we may want to use server-side cookies instead.


suzubara:
Closing this in favor of the actual implementation: https://github.com/USSF-ORBIT/ussf-portal-client/pull/174

PR #146, opened 2021-08-18
ci: Add cypress-axe tests and reports

### LINKED ISSUE #57: Add a11y check to Cypress tests
Set up this plugin to run the pa11y command during Cypress tests, and store artifacts in Github:
https://github.com/mfrachet/cypress-audit

abbyoung:
## Description 

This PR adds the `cypress-axe` library for a11y testing in our e2e tests.

*  Adds a11y check using `cypress-axe` to mvp spec
* Adds ability to log violations to terminal
* Adds ability to save violations as json report

Fixes #57 

## Review Notes
You can test the terminal output locally with `yarn cypress run`. You can also check out the cypress action on this PR and view the artifacts.


cypress/support/commands.js
```diff
@@ -23,3 +23,4 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+import 'cypress-axe'
```
suzubara, line 26: Since `cypress-axe` is imported to `support/index.js` (which also imports this file) does it need to be here at all?

suzubara (approved):
this looks awesome!! I left one small question/comment

cypress/support/commands.js
abbyoung: you're right -- seems to work fine without. pushing up the change now!

suzubara (approved):

PR #148, opened 2021-08-19
ci: Test Storybook builds during CI

### LINKED ISSUE #147: Test Storybook builds in CI
Add a workflow or step to existing workflow in CI that builds the Storybook app to verify it is not broken.

suzubara:
## Description 

While working on another issue, I noticed that `yarn storybook` was breaking for a couple of reasons. This PR both:
*  Fixes the Storybook build (by downgrading certain packages that SB is not compatible with yet) and updating the usage of `storybook-addon-next-router`
* Adds a Storybook build to our existing CI so that in the future, changes that break SB will be caught before being merged

This PR does not persist the Storybook output anywhere, but that would be somewhat trivial to do in the future if we want (for example to [Github pages](https://github.com/storybookjs/storybook-deployer))

Fixes #147 

## Review Notes

If you look at the [Node CI actions for this branch](https://github.com/USSF-ORBIT/ussf-portal-client/actions/workflows/node-ci.yml?query=branch%3A147-storybook-ci) you can see the following:
- This run failed because of the Storybook errors: https://github.com/USSF-ORBIT/ussf-portal-client/runs/3372189706?check_suite_focus=true
- This run passed because the Storybook errors were fixed: https://github.com/USSF-ORBIT/ussf-portal-client/runs/3372363722?check_suite_focus=true

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -39,3 +39,6 @@ jobs:
 
       - name: Run Jest tests
         run: yarn test --coverage
+
+      - name: Build Storybook
+        run: yarn storybook --ci
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -41,4 +41,4 @@ jobs:
         run: yarn test --coverage
 
       - name: Build Storybook
-        run: yarn storybook --ci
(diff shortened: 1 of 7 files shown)


abbyoung (approved):
thank you so much for adding this, and for catching the pkg errors!

PR #150, opened 2021-08-20
test: Add jest-axe library

### LINKED ISSUE #125: Set up jest-axe

abbyoung:
## Description 

This PR adds the jest-axe library to use for a11y testing in unit tests as needed.

Most likely, the majority of automated testing will be done with Cypress/e2e testing, since the context is important. But it may be helpful to test a11y with jest when building certain components. It's a good tool to have ready!

*  Adds jest-axe library
* Adds usage example to `CovidSiteAlert` component

Fixes #125 

## Review Notes
`yarn test` to run unit tests

You can view the docs and example for RTL here: https://github.com/nickcolley/jest-axe#testing-react-with-react-testing-library


src/components/MVP/CovidSiteAlert/CovidSiteAlert.test.tsx
```diff
@@ -3,12 +3,12 @@
  */
 import React from 'react'
 import { render, screen } from '@testing-library/react'
-
+import { axe, toHaveNoViolations } from 'jest-axe'
 import CovidSiteAlert from './CovidSiteAlert'
 
+expect.extend(toHaveNoViolations)
```
suzubara, line 9: I think we can call `expect.extend` in the jest setup file to extend it globally so we don't need to add this to each test file!

suzubara (approved):
minor suggestion but looks great!

src/components/MVP/CovidSiteAlert/CovidSiteAlert.test.tsx
abbyoung: ahh i was trying to get that to work yesterday but couldn't access `expect` in the file

suzubara: yeah, eslint gets mad but I think the tests will still run. I think there is some way of telling eslint that `expect` is a global but I am not sure

abbyoung: oh -- i see. it actually _does_ work, even though eslint is mad that expect is undefined.

suzubara (approved):

PR #183, opened 2021-09-01
refactor(storybook): Organize MVP components in Storybook

suzubara:
## Description 

This PR moves existing MVP components into a **MVP** section in Storybook.

image


next-env.d.ts
```diff
@@ -1,3 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
+
```
abbyoung, line 4: Q about this comment: Should it be here permanently in the file? I noticed working on my last PR that it kept trying to add it, and I thought it was because it was somehow editing it. If we leave it in, and it quits yelling, that's great!

abbyoung (approved):

next-env.d.ts
suzubara: yes! I'm not sure, this might have been added in a `next` update, but it will auto update this file when we run next and changes should be comitted

PR #214, opened 2021-09-08
dev: Header component

### LINKED ISSUE #194: Header component
Create a header component with static, non-functional links to use for beta testing, and add examples to Storybook.

Ref: https://trussworks.github.io/react-uswds/?path=/story/components-header--basic-header

Design:
image

suzubara:
## Description 

This PR adds a Header UI component with static, non-functional links for beta testing. I also extended the existing Logo component to display an abbreviated version.

Fixes #194 

## Screenshots

image


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,56 @@
...
+import LinkTo from 'components/util/LinkTo/LinkTo'
+
+const Header = () => {
+  const [expanded, setExpanded] = useState(false)
+  const handleNavButtonClick = (): void =>
+    setExpanded((prevExpanded) => !prevExpanded)
+
+  const navItems = [
+    <NavLink href="#" key="nav_about">
+      <span>About Space Force</span>
```
abbyoung, line 22: Question for my own code: Is wrapping the text inside of a NavLink in a span best practice? Or is it for additional styling? Would that go for <LinkTo> or other similar link components?

abbyoung (approved):

src/components/Header/Header.tsx
suzubara: OH this is purely because this is how the USWDS Header styles top-level nav links (see the [source code](https://designsystem.digital.gov/components/header/)). that is literally the only reason, so no need to worry about it in other contexts.

PR #215, opened 2021-09-08
dev: Footer component

### LINKED ISSUE #195: Footer component
Create a footer component with static, non-functional links to use for beta testing, and add examples to Storybook.

Ref: https://trussworks.github.io/react-uswds/?path=/story/components-footer--slim-footer

Design:
image

suzubara:
## Description 

This PR adds a Footer UI component with static, non-functional links for beta testing.

Fixes #195 

## Screenshots

image


suzubara:
Hold on https://github.com/USSF-ORBIT/ussf-portal-client/pull/214 to merge

src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,25 @@
+@import 'styles/uswdsDependencies';
+
+.footer {
+  background: #333;
+  color: #fff;
+  @include u-font-size('body', '2xs');
```
abbyoung, line 6: I never know when to use '2xs' or something similar with this utility and when to use another value, e.g. integer. How much does it matter?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import {
+  Footer as USWDSFooter,
```
abbyoung, line 3: oh i have to remember this -- i have some naming collisions with search and this would solve it

abbyoung (approved):

src/components/Footer/Footer.module.scss
suzubara: so usually I'll see what the `px` value is supposed to be in the design, and then look up the corresponding closest value in USWDS. these predefined values are referred to as [design tokens](https://designsystem.digital.gov/design-tokens/) and in this case there's a list of all the font size values at https://designsystem.digital.gov/design-tokens/typesetting/font-size/

abbyoung: wonderful! thank you! i've been doing something similar but think i was looking at system tokens rather than theme tokens.

PR #218, opened 2021-09-08
dev: Static Global Search component

### LINKED ISSUE #198: Global search component
Create a static, non-functional search component to use for beta testing, and add examples to Storybook.

Ref: https://trussworks.github.io/react-uswds/?path=/story/components-search--big-search

Design: 
image

abbyoung:
## Description 
This PR adds a placeholder static search component to fill out the beta layout.

*  Adds Search component
*  Adds tests
*  Static - no working functionality.

Fixes #198 

## Review Notes
View locally using `yarn storybook` or at https://198-global-search-component-static--ussf-storybook.netlify.app

Images aren't being copied properly in the Netlify deploy, so the dropdown arrow is missing on the deployed branch. Going to look into what config needs to change for the deploy.

Screen Shot 2021-09-08 at 4 10 54 PM


src/components/Search/Search.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import styles from './Search.module.scss'
+import LinkTo from 'components/util/LinkTo/LinkTo'
+
+const Search = () => {
+  return (
+    <div className={styles.search}>
+      <form className="usa-search usa-search--big" role="search">
+        <select className="usa-select usa-button" name="options" id="options">
```
suzubara, line 9: both form inputs should have a label element (that can be `.usa-sr-only`) in order to be accessible!

suzubara (commented):
left one comment, looks good otherwise!

src/components/Search/Search.tsx
abbyoung: THANKS!

suzubara (approved):

PR #243, opened 2021-09-27
feat: Delete link from custom collection

### LINKED ISSUE #232: Delete link from Collection
Add a delete handler to the Bookmarks displayed on the Beta landing page that:
- Displays the "Undo" state for a defined period of time (maybe 10s?)
- If the user clicks "Undo", re-displays the Bookmark
- If the user does not click "Undo", the Undo state disappears and the Bookmark remains hidden

image

image

image

suzubara:
## Description 

This PR hooks up the UI for removing a bookmark from a collection. It introduces the `CustomCollection` component, which renders existing Collection & Bookmark components and wraps them with custom functionality (for now, the delete button, but can be used to add/edit in the future). The delete handler first shows an Undo state for 3 seconds, and if the user does not click Undo, the bookmark will be removed (this is happening via the local cache for now).

I also did some minor cleanup in various places

**Outstanding:** the Jest test for deleting a bookmark is skipped, something is not working related to the `@client` directive? But it is passing in Cypress which is more important for now IMO...

Fixes #232 

## Review Notes

* To test, go to the [deployed branch](https://232-delete-link--ussf-portal-client.netlify.app/joinbeta) and try clicking on the "X" buttons in each link in the Example Collection.
* If you then click "Undo" within 3 seconds, the state will revert back. Otherwise, the link should disappear from the collection.
* In Storybook: clicking the delete button will toggle the Undo state, but not actually remove the Bookmark
* Is 3 seconds a good amount of time for the Undo state? Too long?


abbyoung (approved):
🎉 Looks great, thank you for all the additional clean up too.

abbyoung (approved):

PR #248, opened 2021-09-30
[WIP] Docker

suzubara:
## Description 

...

*  
* 
* 

Fixes #


suzubara:
@esacteksab I think switching to distroless is going to take a sec to figure out so I did not include that in this PR. I think we should handle it as a separate issue.

PR #255, opened 2021-10-07
feat: Add existing collections to My Space

### LINKED ISSUE #235: Add existing Collection(s) to My Space
Add a "Select multiple collections" button to the Sites & Application page that toggles between the existing (default) state, and a multi-select state. When in the multi-select state, each Collection should become selectable with a hover state. The user should be able to select 1 or more Collections, and click "Add Selected" to add them to My Space (or "Cancel" to go back to the default state).

image

image

suzubara:
## Description 

This PR adds the UX flow to select existing collections from the Sites & Applications page, and add them to a user's My Space. This includes:
*  A "select" state on the Sites & Applications page that renders each collection as selectable buttons, and any links contained within are not clickable.
* A mutation to add multiple collections in a single operation

Fixes #235 

## Review Notes

* Even though it wasn't specified in the design, I added an "unselect" handler if you click a collection that has already been selected.
* Since the collections are only stored in the local cache, refreshing the page will reset any changes you've made.
* One thing this PR does _not_ implement is storing references to CMS-managed collections -- it simply copies the data from a collection into a user's customized My Space. This means if a user saves a collection and then changes to that collection are made in the CMS, they will not be changed in the user's version as well. Implementing that will require some additional architectural work to determine how the CMS & app databases/APIs might interact. I will create follow up issues for that work.
* **Question:** What should the UX be after saving the selected collections to My Space?
  * Answered: navigate to My Space

## Screenshots

https://user-images.githubusercontent.com/2723066/136426301-18bfb318-bbd2-4fab-9316-edd003031828.mp4

### DIFF
--- a/cypress/integration/sites-and-applications.spec.js
+++ b/cypress/integration/sites-and-applications.spec.js
@@ -21,6 +21,40 @@ describe('Sites and Applications', () => {
     cy.contains('Application name').should('not.exist')
   })
 
+  it('can add collections from the Sites & Applications page to My Space', () => {
+    cy.contains('My Space')
+    cy.contains('Career').should('not.exist')
+    cy.contains('Medical & Dental').should('not.exist')
+    cy.contains('Life & Fitness').should('not.exist')
+
+    // Go to Sites & Applications
+    cy.contains('All sites & applications').click()
+
+    // Enter select mode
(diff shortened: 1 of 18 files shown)


jbecker01 (approved):
Works like a charm!

PR #260, opened 2021-10-08
feat: Add new section to My Space

### LINKED ISSUE #236: Add Widget UX
Create an empty Widget with a dropdown menu with one item, "Select collection", that brings the user to the Sites & Applications page in multi-select mode. This Widget should be the last item displayed in the My Space component.

image

suzubara:
## Description 

This PR adds a button to the end of My Space that can be used to add new sections. Right now the dropdown only contains one option, which brings the user to the Sites & Applications page with the select mode turned on.

I also made a change to the DropdownMenu component in order to use more flexible elements for the toggle button.

Fixes #236 

## Review Notes

* From the My Space page, click on "Add Section" and add existing collections. Verify you are brought to Sites & Applications where you can add sections to My Space.

addWidget


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -1,13 +1,16 @@
...
-  menuIcon: React.ReactNode
-  ariaLabel: string
-  onMenuClick: () => void
+  menuIcon?: React.ReactNode
+  ariaLabel?: string
+  onMenuClick?: () => void
   isActive: boolean
-  dropdownRef: React.RefObject<HTMLDivElement>
+  dropdownRef?: React.RefObject<HTMLDivElement>
+  triggerEl?: React.ReactNode
```
suzubara, line 13: @abbyoung I made some changes to the DropdownMenu to allow passing in an entire element to use as the open/close toggle, because I needed to be able to style this button and make sure it had certain attributes (like `id`). This leaves the prop types in a weird state because if someone passes in a `triggerEl` that means they don't need to pass in props like `menuIcon, ariaLabel, onMenuClick`... we can either update this component so that it's compatible with either method, _or_ require that `triggerEl` is always passed in. What do you think?

esacteksab:
It does what it says on the package, but I have feelings. :shipit:

src/components/DropdownMenu/DropdownMenu.tsx
abbyoung: i know we briefly discussed in person but for the record: i like toggleEl much better 👍

PR #279, opened 2021-10-19
feat: Added typography styles

### LINKED ISSUE #285: Implement type styles for headers & body text
Add documentation & examples in Storybook of what the default type styles should be for header (h1-h6) elements and body (p) elements.

Apply these styles to the new beta pages.

malworks:
## Description 
Added the current typography (h1-h6, body) to storybook.
Added typography globally to the application

Fixes #285 

## Review Notes
- h1-h6 and p should be Trebuchet (and correct sizing found in Figma)


suzubara:
@malworks is this branch up to date with your latest? I don't see the changes where these get applied to the application code. Also you can hit "Update branch" on the PR to pull in the latest code from `main`

malworks:
@suzubara Done!

src/layout/Beta/DefaultLayout/DefaultLayout.tsx
```diff
@@ -30,7 +30,7 @@ const DefaultLayout = ({ children }: { children: React.ReactNode }) => {
       <a className="usa-skipnav" href="#main-content">
         Skip to main content
       </a>
-      <div className={styles.siteContainer}>
+      <div className={`${styles.siteContainer}, ${styles.defaultTheme}`}>
```
suzubara, line 33: There shouldn't be a comma between these 2 classes, this is affecting the siteContainer class not being applied correctly. apologies I might have typed this incorrectly when we were pairing!

suzubara (commented):
Nice! my main comment is that we should move the `Typography.stories.tsx` file out of the `components` directory, because typography is not a component on its own. Can we move that file to `/src/stories/typography.tsx` (lowercase)? I think we need to create the `src/stories` directory to use for Storybook files that are demonstrating global styles. We can also delete the `Typography.module.scss` file altogether & apply styles directly inside of the stories file.

Also, I have a question about the new type styles: should "Trebuchet" be the default font across the board? if so then I think we can add CSS declarations to reflect that. Should the `body1` style be applied to `p` tags? There are also some lingering header styles in some module files that are conflicting with the new styles so we may want to remove those for consistency.

@malworks how would you prefer help with the above? I'm happy to pair, leave it up to you, or push some changes to this branch myself. just let me know!

src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const Typography = () => {
+  return (
+    <div className="">
+      <h1 className="">Typography</h1>
+      <p className="">Typeface: Trebuchet</p>
+      <table className={styles.typeTable}>
+        <tr>
+          <td>
+            <h6>42px / 2.625rem / 0</h6>
```
suzubara, line 17: what does the `0` here indicate?

suzubara:
I added https://github.com/USSF-ORBIT/ussf-portal-client/issues/285 to help track this work

PR #288, opened 2021-10-25
build: Add mongo and mongo-express to Dockerfile.dev

### LINKED ISSUE #272: Docker compose for client app & other services (for development)
Set up docker compose file to be used for local development only, that starts the portal client app in dev mode, and also a MongoDB service

### LINKED ISSUE #273: Convert initCache to seed data for local development
Instead of initializing the local ApolloClient cache with seed data, we should be able to run a script that seeds the data in our database. This will need to include creating a single user so custom collections/bookmarks can be associated with that user.

abbyoung:
## Description 

This PR builds on the Dockerfile.dev for local development by adding a MongoDB service and (temporary?) GUI, Mongo Express.

* Adds MongoDB service and initializes a database named `dev` with a `users` collection and one document.
* Adds Mongo Express for quick testing / verification of database. (localhost:8888)

## Testing

**On first run:**

* Pull down the branch and run `yarn install`
* `docker compose up`
* You should be able to view the app on localhost:3000, and Mongo Express on localhost:8888
* Mongo Express should show a table named `dev` with a `users` collection and one document
* When finished, run `docker compose down`

**On subsequent runs:**

* If you want to **re-initialize the database** by running the `mongo-init.js` script, you must remove the mongo data volume first: `docker volume rm ussf-portal-client_mongodb_data_container`

When in doubt, `docker prune` all the things and try again.

Fixes #272
Fixes #273 

## Review Notes
* ~Since this Dockerfile is only for local development, the not-so-real db credentials are kept in the Dockerfile. If y'all would like to move those vars (or any others) out into a local env file, let me know.~
* You can also connect to the db using [MongoDB Compass](https://www.mongodb.com/products/compass). 
  * Connection string: `mongodb://127.0.0.1:27017/?readPreference=primary&appname=MongoDB%20Compass&ssl=false`

## To Do
- [x] Update some docs somewhere, probably


docker-compose.yml
```diff
@@ -13,3 +13,36 @@ services:
...
       - '3000:3000'
+    environment:
+      - MONGO_URL=mongodb://mongo:27017
+      - MONGODB_DB=dev
+    stdin_open: true
+    depends_on:
+      - mongo
+  mongo:
+    container_name: mongo
+    image: mongo:5.0.3
```
esacteksab, line 24: This is too new and we risk incompatibility with DocumentDB

Need 3.6 or 4.0 https://aws.amazon.com/documentdb/faqs/ 
https://docs.aws.amazon.com/documentdb/latest/developerguide/mongo-apis.html

abbyoung: we'll be using 4

docker-compose.yml
```diff
@@ -13,3 +13,30 @@ services:
       - /app/.next
     ports:
       - '3000:3000'
+    environment:
+      - MONGO_URL=mongodb://mongo:27017
+      - MONGODB_DB=dev
```
suzubara: super tiny nit: `MONGO_DB_NAME` might be a little clearer that this is the name of the database?

docs/development.md
```diff
@@ -67,14 +67,70 @@ You can start the NextJS server in "production mode" (i.e., the server will run
...
+docker volume rm ussf-portal-client_mongodb_data_container
+
+docker compose up
+```
+
+### Known limitations
+
+- Currently Docker is not set up to run the static site.
+
+- If a change is made to package.json, you'll need to shut down the environment and rebuild the app image using `docker compose up --build`
```
suzubara, line 141: Also small technical change: `If a change is made to package.json dependencies (node_modules)`

since you can make changes in package.json that are not dependency-related and shouldn't require a rebuild

mongo-init.js
```diff
@@ -0,0 +1,65 @@
+/* eslint no-undef: "off" */
```
suzubara, line 1: I am assuming this script should only run when initializing the local dev instance right? Is there some way we can clearly delineate that?

package.json
```diff
@@ -37,6 +37,7 @@
     "@trussworks/react-uswds": "2.4.1",
     "classnames": "2.3.1",
     "graphql": "16.0.1",
+    "mongodb": "^3.5.9",
```
suzubara, line 40: If you pre-emptively pin this version to the one that's actually installed (looks like `3.7.3` according to the yarn.lock file) then Renovate bot won't open a PR to pin it

suzubara (approved):
ALL of my comments are very small, optional, and could easily be applied in future PRs!!! :shipit:

PR #296, opened 2021-10-27
dev: Set up SAML authentication flow with local test IdP

### LINKED ISSUE #289: Integrate portal with SAML library and test IdP
- Add test SAML IdP (https://hub.docker.com/r/kristophjunge/test-saml-idp/) to docker compose with a custom user config file for local development
- Add basic SAML proof of concept to portal app (login URL and login callback)
  - Possible SAML lib options are next-auth or passport-saml
  - Examples: https://github.com/Jenyus-Org/next-auth-saml
  - http://www.passportjs.org/packages/passport-saml/

AC: User can click a button to login (initiate a SAML request), and upon a successful callback a session is created for them.

suzubara:
## Description 

This PR sets up a very rudimentary SAML Service Provider in the portal application, as well as a test IdP (Identity Provider) for use in local development.

### IdP service
* Running `docker-compose up` will now also start the `idp` service on ports 8080, 8443 (uses [kristophjunge/docker-test-saml-idp](https://github.com/kristophjunge/docker-test-saml-idp))
* You can access the IdP admin UI at http://localhost:8080/simplesaml
* The IdP includes a single test user (`user1:user1pass`) but more can be added in `users.php`
* This IdP will only ever be used in local development and should never be used in a deployed environment

### SAML (SP) implementation

As part of this work I evaluated two options for implementing SAML in the portal application: **[passport-saml](https://github.com/node-saml/passport-saml)** and **[next-auth](https://github.com/nextauthjs/next-auth)**. Below is my impression of the two approaches:

#### **[passport-saml](https://github.com/node-saml/passport-saml)**

* A provider built for [Passport](http://www.passportjs.org/), a commonly used authentication library for NodeJS apps
* Since Passport is middleware, which NextJS does not support out of the box, this approach also requires adding [next-connect](https://github.com/hoangvvo/next-connect) to expose middleware handlers in server-side code. While I'd normally prefer to avoid this kind of overhead, a tool like this will be useful for other things as our app grows, for example adding common handlers for API routes such as for authorization and error handling
* Unlike Next-auth, Passport _only_ handles the authentication mechanism, so it will still be up to us to manage user sessions
* I also added **[passport-saml-metadata](https://github.com/compwright/passport-saml-metadata)**, a utility helper for getting IdP configuration (including certs) from metadata XML. (I don't know if an equivalent helper exists for `saml2-js` but I also didn't look very hard). This will let us change IdP settings just based on the XML path, and avoid having to install and rotate certs ourselves.
 
#### **[next-auth](https://github.com/nextauthjs/next-auth)**
* Next-auth is an all-in-one authentication solution for NextJS applications. This was a tempting option since it is specifically built to work with NextJS and wouldn't require adding a library for injecting middleware.
* Next-auth does _not_ have a SAML implementation out of the box. I specifically investigated [this example](https://github.com/Jenyus-Org/next-auth-saml) which creates a custom credentials provider using [saml2-js](https://github.com/Clever/saml2).
* Ultimately I decided to move forward with `passport-saml` instead for the following reasons:
  * `next-auth` looks to be more of an all-encompassing solution for authentication in NextJS apps -- it is also responsible for managing sessions in a database and defining the authentication API routes. It seems to be designed as a one-size-fits-all for applications that may be supporting multiple or more traditional log in methods (such as OAuth or username/password), which frankly seems like overkill for our purposes, since we only need to implement SAML (for now).
  * It requires a CSRF token in POST requests, which isn't present in a SAML response. The example code I linked above gets around this by generating an HTML form and submitting it as part of the response handler, which felt hacky and unnecessary.
  * Since Next-auth handles sessions it requires either configuring a database, or storing data in a JWT. Next-auth does not have a database adapter for Redis, so we would need to build that ourselves.

While Next-auth _would_ do more for us than Passport, I'm inclined to stick with Passport at least for now. I believe it will give us more fine-grained control over our authentication flows and handling user data, and right now we don't have the use case of needing to integrate with multiple authentication strategies. There is no reason why we could not migrate from Passport to Next-auth in the future, since the configuration details of the SAML service provider wouldn't change. But I think for now, as we are still feeling out the needs and architecture of this application, it'll be simpler to use Passport.

Fixes #289 

## Review Notes

To test:
- Check out this branch locally and run `yarn install`
- Start the services with `docker-compose up -d`
- Navigate to http://localhost:8080/simplesaml to verify the test IdP is running
- Navigate to the portal app login API route: http://localhost:3000/api/auth/login
- You should be redirected to a login form on the IdP service (localhost:8080)
- Enter the credentials for the test user (`user1`, `user1pass`)
- You should be redirected back to the portal app (http://localhost:3000/api/auth/login) and see test user data displayed
  - i.e., `Logged in as user: Logged in as user: 1540897553 TEST.USER.1234567890@testusers.cce.af.mil AFIN_TRANSIT,AF_USERS`
- Navigate to the logout API route: http://localhost:3000/api/auth/logout
- You should see the message that you've been logged out
- If you go back to http://localhost:3000/api/auth/login you should be redirected back to the IdP login form, verifying you have indeed been logged out


suzubara:
I started writing Cypress tests for login/logout against the test IDP, but ran into issues because Cypress [restricts navigating to more than one origin in a single test](https://docs.cypress.io/guides/guides/web-security). You can run the Cypress `saml.spec.js` test locally and view it working, but Cypress reports it as a failure because of the origin redirects.

Since the complete flow is still a WIP, and I've already started making some refinements in a [follow up PR](https://github.com/USSF-ORBIT/ussf-portal-client/pull/331), I'd prefer to go ahead & merge this in, and complete our tests and documentation once we have a better handle of the end-to-end flow.

abbyoung (approved):

suzubara:
@abbyoung I thought I would be clever and add some tests but then they didn't work 😭  I think building out a mock Passport/SAML implementation is going to be a whole thing so I am skipping them for now

abbyoung (approved):

PR #293, opened 2021-10-25
feat: Sites & Applications UX fixes

suzubara:
## Description 

This PR makes some small UX/design fixes as described in #286 (everything listed in the **COMPLETED** section). It does _not_ cover the items contained in separate issues.

- [x] Adjust MySpace columns to 3, maybe adjust margins/padding
- [x] Change color on search dropdown text to white 
- [x] Change padding on collections boxes (24px for top/bottom and 16pc for L/R)
- [x] Even out padding within myspace container (24px on all sides)
- [x] Remove side nav links that are unused (My profile, Subordinate profiles, etc...)
- [x] Change all footer text to white
- [x] Add hover states to buttons
- [x] Remove hover state for search dropdown
- [x] "Add section" box height should be equal to a collection card with 5 links inside it
- [x] "Search" text is not using trebuchet
- [x] In "alphabetical" view, description columns will need to be moved to the left to make room for more content (Note: this might actually be resolved on it's own once content is added)
- [x] Add hover states to main nav
- [x] Update button styles on alphabetical/type sorts
- [x] ReactUSWDS ComboBox scroll bug (https://github.com/trussworks/react-uswds/issues/1693)

Fixes #286 

## Review Notes

Review the deployed app & Storybook against the list of UX fixes above and verify that they are in fact completed.

PR #301, opened 2021-10-29
feat: Validate custom URL when adding a link to a collection

### LINKED ISSUE #270: Validate URL when user enters a custom link
When adding a custom link (bookmark) to a collection, we should validate that the URL entered is valid.

suzubara:
## Description 

This PR adds native form validation to the Add Link input, to verify that when a user is entering a custom link they provide a valid, complete URL. Since this is an initial implementation of this feature, it just uses a regex to verify the input. We may want to use a more robust method in the future since using regex to validate URL strings is usually not recommended.

Fixes #270 

## Review Notes

- Go to My Space and click "+ Add link" under a collection
- Try entering a non-URL into the input (i.e., www.example.com). You should see an error message and be prevented from adding the link.
- Verify you can enter a valid URL into the input (i.e., http://www.example.com) and proceed.
- Verify you can also still select an existing CMS link and add it to the collection.

image

### DIFF
--- a/cypress/integration/sites-and-applications.spec.js
+++ b/cypress/integration/sites-and-applications.spec.js
@@ -192,21 +192,20 @@ describe('Sites and Applications', () => {
           .then(($el) => $el[0].checkValidity())
           .should('be.false')
 
-        /*
-          // TODO - URL validation
         cy.findByLabelText('URL')
           .type('not a URL')
           .then(($el) => $el[0].checkValidity())
           .should('be.false')
-          */
 
         cy.findByLabelText('URL')
           .clear()
-          .type('http://www.example.com')
+          .type('http://www.example.com{enter}')
+          .blur()
           .then(($el) => $el[0].checkValidity())
           .should('be.true')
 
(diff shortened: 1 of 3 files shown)


suzubara:
hold on https://github.com/USSF-ORBIT/ussf-portal-client/pull/293

jbecker01 (commented):
@suzubara 
if I enter in a non-valid URL, delete it, and then select a predefined URL from the dropdown menu, I still get the error message when clicking on the Add Site button.

Screen Shot 2021-10-29 at 1 43 31 PM

suzubara:
This will require https://github.com/trussworks/react-uswds/issues/1730 to fix

suzubara:
@jbecker01 Never mind, I think I got around it. Can you check again?

suzubara:
@jbecker01 Never mind I think this is fixed now! Do you mind checking again?

malworks (approved):
I couldn't recreate John's issue. I think it's fixed!

jbecker01 (approved):
You fixed it!

abbyoung (approved):

PR #303, opened 2021-11-01
ci: Adding AWS Actions

### LINKED ISSUE #308: Build Docker image and publish to ECR in AWS Dev alongside Artifactory in C1 Dev/Int

esacteksab:
## Description 

`dev-aws-ecr.yml` will run when merged to main branch. It will get the tag. Log into ECR. Build the image, tag the image based on the git tag and push the image to the Elastic Container Registry in our Dev AWS Account. It will then download the task definition for the dev stack, update it with the image created in the prior step, then update the service with the new image, deploying it. It will then upload the image to C1 Artifactory instance in dev.

`prototype-aws-ecr.yml` is similar to `dev-aws-ecr.yml` but rather than running on merge to main, it will run when a tag of `release-2.0.0-alpha.*` is created. It will not upload to C1 Artifactory. The workflow is otherwise the same in build and deployment deploying to the prototype stack.

Fixes #308

### DIFF
--- a/.github/workflows/dev-aws-ecr.yml
+++ b/.github/workflows/dev-aws-ecr.yml
@@ -19,6 +19,7 @@ jobs:
       run : |
         echo $RELEASE_VERSION
         echo ${{ env.RELEASE_VERSION }}
+        echo ${{ github.ref }}
     # - name: Configure AWS credentials
     #   uses: aws-actions/configure-aws-credentials@v1
     #   with:
--- a/.github/workflows/dev-aws-ecr.yml
+++ b/.github/workflows/dev-aws-ecr.yml
@@ -14,11 +14,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Get tag
-      run: echo "RELEASE_VERSION=${GITHUB_REF#REFS/*/}" >> $GITHUB_ENV
+      run: echo "RELEASE_VERSION=${GITHUB_REF:10}" >> $GITHUB_ENV
     - name: test
       run : |
         echo $RELEASE_VERSION
-        echo ${{ env.RELEASE_VERSION:10 }}
+        echo ${{ env.RELEASE_VERSION }}
     # - name: Configure AWS credentials
     #   uses: aws-actions/configure-aws-credentials@v1
     #   with:
--- a/.github/workflows/dev-aws-ecr.yml
+++ b/.github/workflows/dev-aws-ecr.yml
@@ -1,6 +1,12 @@
+# on:
+#   release:
+#     types: [published]
+
 on:
-  release:
-    types: [published]
+  push:
+    branches: [main]
+  pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/dev-aws-ecr.yml
```diff
@@ -0,0 +1,69 @@
+name: Docker Stuff
```
suzubara, line 1: 😂  do you mind giving these a little more descriptive names? perhaps `Docker build main`, `Docker build alpha`?

.github/workflows/dev-aws-ecr.yml
```diff
@@ -0,0 +1,69 @@
...
+  push:
+    branches: [main]
+
+jobs:
+  build-and-push-aws:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Get tag
+      run: echo "RELEASE_VERSION=${GITHUB_REF:10}" >> $GITHUB_ENV
```
suzubara, line 13: what does the `:10` part of the `GITHUB_REF:10` var mean?

.github/workflows/dev-aws-ecr.yml
```diff
@@ -0,0 +1,69 @@
...
+        task-definition: dev-task-definition.json
+        container-name: portal-client
+        image: ${{ steps.build-image.outputs.image }}
+
+    - name: Deploy Amazon ECS task definition
+      uses: aws-actions/amazon-ecs-deploy-task-definition@v1
+      with:
+        task-definition: ${{ steps.task-def.outputs.task-definition }}
+        service: portal-client
+        cluster: app-my-webapp-dev
```
suzubara, line 55: is this cluster name right? the prototype one seems more descriptive

suzubara (approved):
exciting!! I left a couple of questions but it is probably good! 😅

.github/workflows/dev-aws-ecr.yml
esacteksab: I don't mind at all!! Good idea!

.github/workflows/dev-aws-ecr.yml
esacteksab: It takes the first 10 characters of `GITHUB_REF`. I just want `vx.x.x` rather than `10chars/vx.x.x`

.github/workflows/dev-aws-ecr.yml
esacteksab: It _was_ until I rebuilt last night. Good catch! Let me fix!

suzubara (approved):
🎉

.github/workflows/dev-aws-ecr.yml
noahfirth: Hahahaha, @esacteksab, you name things the exact same way I do!

PR #327, opened 2021-11-04
fix: Production Dockerfile

suzubara:
## Description 

Updates to the production Dockerfile: using `slim` instead of alpine/distroless because of Keystone/Prisma (we'll come back to this), and fix passing env variables for building.

To test:
```
docker build -t <image tag> . --build-arg NEXT_PUBLIC_SITE_URL=${NEXT_PUBLIC_SITE_URL}
docker run -p 3000:3000 <image tag>
```

There should be no server errors in the console about Prisma when viewing My Space or Sites and Applications.


esacteksab (approved):

Dockerfile
```diff
@@ -1,52 +1,38 @@
...
-COPY . .
-COPY --from=deps /app/node_modules ./node_modules
-
-# Build the app
 ENV NEXT_TELEMETRY_DISABLED 1
-RUN yarn build
 
-# Install production deps only
-RUN rm -rf node_modules && yarn install --production --ignore-scripts --prefer-offline
+ARG NEXT_PUBLIC_SITE_URL
```
esacteksab, line 16: Do we need to set these in Github Actions where it will build the container?

suzubara: Yes we'll always want to set them

esacteksab (approved):

PR #340, opened 2021-11-05
chore(deps): Security updates

suzubara:
## Description 

This updates as many packages as I could that did not require code changes, and also updates yarn resolutions to resolve some of the recent security vulnerabilities that were flagged.

Not all vulnerabilities have been resolved, but all remaining are linked to devDependencies (such as Storybook) and should not introduce vulnerabilities into a production system (see [this article](https://overreacted.io/npm-audit-broken-by-design/) for more info).

## Review Notes

- All tests should pass and you should be able to run yarn scripts locally without errors
- `yarn audit --groups dependencies` should return 0 vulnerabilities (this limits checks to production deps only)


esacteksab (approved):
THANK YOU!

PR #331, opened 2021-11-05
feat: SAML integration

suzubara:
## Description 

This PR adds some Docker config & documentation around how to test SAML authentication against the actual provider. I also cleaned up some env vars that are not being used.

In the future it would be nice to use chamber or something to store sensitive env vars & our certificates so someone setting up their local environment for SAML doesn't have to ask someone for those values.

Fixes #291 

## Review Notes

- Follow the instructions to test our real SAML IdP locally and verify you are able to log in with your CAC and view real data.
- Test logging into our dev environment with your CAC:
    - Navigate to https://dev.ussforbit.us/api/auth/user & verify you see `{"user":null}`
    - Go to https://dev.ussforbit.us/api/auth/login follow the prompts to log in with your CAC and finally you should end up seeing your CAC data in JSON (with your real name)
    - Verify the session data persists by returning to https://dev.ussforbit.us/api/auth/user and you should see the same JSON data
    - Log out by going to https://dev.ussforbit.us/api/auth/logout (nothing will appear on the page)
    - Verify you deleted the session by returning to https://dev.ussforbit.us/api/auth/user & verify you see `{"user":null}` again

PR #349, opened 2021-11-09
feat: Store authenticated sessions in redis

### LINKED ISSUE #290: Stand up redis for storing session data
- Add redis to docker compose for local development
- Use redis for reading & writing session data

suzubara:
## Description 

This PR adds session handling to the portal app, storing session data in Redis, and makes the authentication API endpoints more consistent.

* Configures a redis client & adds environment var `REDIS_URL`
* Adds `redis` service to docker-compose for local development
* Adds unit tests to `/api/auth` endpoints, with a mock implementation of the SAML authentication strategy and uses in-memory sessions instead of Redis
* All `/api/auth` endpoints return JSON instead of redirects (this was easiest to test, but may change based on where https://github.com/USSF-ORBIT/ussf-portal-client/pull/331 ends up)
* Adds `axios` to make SLO POST requests on the server

Fixes #290


src/pages/api/auth/[[...action]].ts
```diff
@@ -1,75 +1,113 @@
...
+    await req.session.destroy()
 
-    passport.logoutSaml(logoutRequest, (err, request) => {
-      if (!err && request) {
-        res.redirect(request)
+    passport.logoutSaml(req, async (err, logoutRequest) => {
+      if (!err && logoutRequest) {
+        // TEMPORARY - DEVELOPMENT ONLY
+        // because our test IDP does not support SLO HTTP-POST bindings
+        if (
```
Comment (suzubara) [inline src/pages/api/auth/[[...action]].ts:81]: I added this because the test IDP docker image we are using does not seem to allow us to configure the logout endpoint to accept HTTP-POST requests, only HTTP-Redirect. However the actual GCDS integration _only_ supports HTTP-POST logout. Long term I found a node-based [test IDP](https://github.com/mcguinness/saml-idp) that might be worth configuring ourselves for development to more closely mirror our actual auth, but that is not a priority right now.

abbyoung (dismissed):
🎉 🚢

abbyoung (approved):

PR #357, opened 2021-11-12
ci: Making package files owned by everyone to allow renovatebot to automerge.

esacteksab:
## Description 

Renovatebot won't automerge because of `CODEOWNERS`, but thanks to a fellow Trussel, and this [link](https://github.com/renovatebot/renovate/issues/6473#issuecomment-855543240), this might do what we want which is make the package files owned by everyone...which says it should allow renovatebot to automerge.

Fixes #


suzubara (commented):
https://docs.renovatebot.com/automerge-configuration/ gives me a 404

PR #361, opened 2021-11-16
ci(docker): Reduce size of Docker production image

suzubara:
## Description 

* Only install production dependencies after building NextJS app
* Use node-slim for final image with deps installed for Prisma
* Ignore `.keystone` directory


abbyoung (approved):
amazing 😭

PR #363, opened 2021-11-17
ci(cypress): Run Cypress CI against docker compose stack

suzubara:
## Description 

This PR makes some modifications to our Cypress config so that e2e tests run against our docker compose stack.

*  We have a new `docker-compose.e2e.yml` that spins up the stack specifically for running e2e tests agains. This is used in Github CI, and uses the production app Dockerfile to build/start the NextJS server. Right now it only spins up the app & mongodb.
* Moved Cypress files & dependencies into a separate `/e2e` directory (which also let us remove some Cypress-only deps from the application `package.json`)
* Updated docs about how to run Cypress tests locally


docker-compose.e2e.yml
```diff
@@ -0,0 +1,34 @@
...
+    environment:
+      MONGO_INITDB_DATABASE: dev
+    volumes:
+      - mongodb_data_container:/data/db
+      - ./mongo-init.js:/docker-entrypoint-initdb.d/mongo-init.js
+    expose:
+      - '27017'
+
+volumes:
+  mongodb_data_container:
```
abbyoung, line 34: i don't believe we need to persist the data in a volume, but i also don't know that it matters

abbyoung (approved):
inspired ✨

PR #366, opened 2021-11-18
chore: Add url descriptions

### LINKED ISSUE #287: Update URLs and their descriptions
Update the existing URLs with a description in the CMS

https://docs.google.com/spreadsheets/d/1Ah0E7tGgrbkL0KOx5dcFL4XCcZh2JDPzTDutfot7MXw/edit?usp=sharing

jbecker01:
## Description 

Updates to URLs and descriptions in the CMS

*  Added URLs and descriptions for all collections
*  Created new collections for Supervisory, Education, Public Military Websites, and Public Social Platforms
*  Added all URLs from AF portal

Fixes #287


noahfirth (dismissed):
Approved! Are we using the description as a tooltip or is that going to show up somewhere else?

suzubara (approved):

suzubara:
> Approved! Are we using the description as a tooltip or is that going to show up somewhere else?

@noahfirth Right now the description only shows up in Alphabetical view. I'm not sure if there's someplace else we want to show them, but we could add them in a [title attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) pretty easily

suzubara (approved):

PR #370, opened 2021-11-18
build: Add endpoint to return build/system info

### LINKED ISSUE #304: Include portal app version in build somewhere (endpoint, source, header...)
Include the built version somewhere in the application so we can verify what version is running in an environment

abbyoung:
## Description 

Adds `/api/sysinfo` endpoint to return build version, node environment, and git SHA.

Fixes #304 

## Review Notes

~|@esacteksab -- @suzubara and I talked through how this might work in deployed envs, passing a build arg in `dev-aws-ecr` to Docker, and then storing that as an env variable. Would love your feedback! Also, my local linter insisted on reformatting the indentation of the file...is it gonna break? 😱~

~Refactored and no longer have to pass anything special into Docker or workflow files 🎉~

Back to passing the image tag into the build ✅


esacteksab (dismissed):

esacteksab (approved):

.github/workflows/dev-aws-ecr.yml
```diff
@@ -8,54 +8,54 @@ jobs:
...
+      - name: Build, tag, and push image to C1 Artifactory
+        id: build-c1-image
+        env:
+          IMAGE_TAG: ${{ github.sha }}
+          C1_REGISTRY: ${{ secrets.C1_REGISTRY }}
+          C1_REPOSITORY: ${{ secrets.C1_REPOSITORY }}
+        run: |
+          sudo sh scripts/add-dod-cas.sh
+          docker login -u portal -p ${{ secrets.C1_ARTIFACTORY_TOKEN }} $C1_REGISTRY
+          docker build -t $C1_REGISTRY/$C1_REPOSITORY:$IMAGE_TAG . --build-arg BUILD=$IMAGE_TAG
```
suzubara, line 58: Does the Dockerfile need to be updated to set this argument as an env var?

abbyoung: that would be helpful! not sure what happened, looks like all my changes didn't get pushed up. thank you!

.github/workflows/dev-aws-ecr.yml
```diff
@@ -35,17 +35,16 @@ jobs:
 
       run: |
         echo $CERT > saml.pem
-        docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG .
+        docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG . --build-arg BUILD=$IMAGE_TAG
```
esacteksab, line 38: In addition to making this change here, you'll also want to make it here https://github.com/USSF-ORBIT/ussf-portal-client/blob/main/.github/workflows/build-push-c1-art.yml (might need to merge main into your branch as this was merged this morning).

esacteksab (dismissed):

.github/workflows/dev-aws-ecr.yml
abbyoung: ahh thank you! added 🌟

suzubara (commented):
I tested this locally by building the docker image, and found I actually needed to define the `BUILD` env var in the environment itself, I think because at the point in the Dockerfile when the build step happens, the `NODE_ENV` is not set to `production` yet. I added the var to my docker compose `environment:` so maybe passing into `env:` in the workflow files will achieve the same?

esacteksab:
I know quite a bit of work has been done on this, but stumbled upon this, this morning https://nextjs.org/docs/api-reference/next.config.js/configuring-the-build-id

suzubara:
2 suggestions:
- Use an alternative method to get the package version, perhaps https://stackoverflow.com/questions/48609931/how-can-i-reference-package-version-in-npm-script/48619640
- Let's use the next generateBuildId to get the SHA instead of running exec code in an API endpoint, also look at https://github.com/nexdrew/next-build-id

suzubara (dismissed):
💯  beautiful

suzubara (approved):

PR #384, opened 2021-11-30
ci: fix gha docker build step

esacteksab:
## Description 

Built off of `291-saml-branch`

Uses GitHub actions to create `saml.pem` in the root of the repository which is put into the container. Then `add-dod-cas.sh` downloads the DoD certs and now constructs a PEM which is used by `NODE_EXTRA_CA_CERTS` -- this is currently on dev.ussforbit.us to test. Screenshots in Slack.


suzubara (approved):
🎉

PR #394, opened 2021-12-01
ci: abstracting build components to work around issues with GHA.

esacteksab:
## Description 

1) We don't need the federation script in C1 step, so removing it from `dev-aws-ecr.yml` GHA and `add-dod-cas.sh` script.
2) Small rearrangement of Dockerfile to condense `RUN` statements and tell apt not to install recommendations. This trims a few MB from the image, but every little bit helps.


.github/workflows/dev-aws-ecr.yml
```diff
@@ -57,7 +57,6 @@ jobs:
         CERT: ${{ secrets.SAML_CERT }}
 
       run: |
-        sudo echo $CERT > /usr/local/share/ca-certificates/federation.dev.cce.af.mil.crt
```
suzubara, line 60: Why isn't this needed in the C1 step? I think I'm confused as to how the docker image differs between this step and the AWS step

esacteksab: Federation use used for SAML endpoint. C1 Artifactory doesn't have anything to do with SAML so this cert isn't necessary. While the C1 step does not need this cert to access Artifactory, it does still need `saml.pem` to exist so it can build the container. Fixing this.

.github/workflows/dev-aws-ecr.yml
```diff
@@ -57,7 +57,7 @@ jobs:
...
+        echo $CERT > saml.pem
         sudo sh scripts/add-dod-cas.sh
```
suzubara, line 61: Does this script run twice, once here and once in the Dockerfile?

suzubara (approved):
nice clean up!

.github/workflows/dev-aws-ecr.yml
esacteksab: Yes. In the Dockerfile `saml.pem` is necessary for creating the GCDS chain. `add-dod-cas.sh` is called in the Dockerfile to also create the chain and overall, just exist for other DoD systems in the future. 

`add-dod-cas.sh` also runs on the Github Runner, because we need it so we can `docker [login/push]` to artifactory which uses a self-signed cert that's issued by DoD CA's.

suzubara (approved):

Issue #379, opened 2021-11-23
Use logged in user for GraphQL context


suzubara:
Implement auth in the GraphQL API:

- Make sure GraphQL resolvers have access to logged in user object
- Verify operations are only performed on the logged in user document
- Return access denied error if user doesn't match

Resources:
- https://www.apollographql.com/docs/apollo-server/security/authentication/
- https://www.apollographql.com/blog/backend/auth/a-guide-to-authentication-in-graphql/

Issue #292, opened 2021-10-25
Implement session timeout


suzubara:
Implement timeouts on both the server & client that call the SAML log out handler (https://github.com/USSF-ORBIT/ussf-portal-client/issues/128) after a pre-determined time period. Session data should be deleted and the user should be redirected to the login page.

PR #395, opened 2021-12-01
feat: User authentication flow

### LINKED ISSUE #126: Handlers & logic for authenticated routes
- Add a placeholder un-authenticated page with a button to initialize a SAML request
- Add logic to all existing beta pages that requires authentication and redirects to the login page if there is no session
- Existing MVP pages should not be affected
- Display user first/last name in Personal Data
- Add Log Out button to Beta pages (https://www.figma.com/file/iPOdDkRxCaeoDoQ4JXjxlk/Space-Force-Service-Portal?node-id=2688%3A30769)

suzubara:
## Description 

This PR adds code to all existing routes (except for `/login` and `/login-notice`) that require the user to be logged in via SAML in order to view. If the user is not logged in, they will be automatically redirected back to the Login page.

Behind the scenes:
* `authContext`: For storing & accessing user/session data in the application state (this state is wiped clean on page load, used so the client doesn't have to re-fetch the user on each client-side navigation)
* `useUser` hook: used to initiate a user prop from SSR _or_ fetch the user client-side and set in the authContext
* `requireAuth`: server-side props function for loading user from Redis - currently not used on any routes, will revisit in #404 
* I had to modify the beta redirects in order to not redirect `/login` to `/beta/login`. This means non-beta URLs are now accessible even while the beta cookie is set, but I think that's ok.
* Updated `/api/auth` endpoints to use redirects & proper statuses

UI:
* Hook up Log in button to initiate SAML login
* Adds log out buttons to navigation on both MVP & Beta
* Render actual user name on Beta header
* Modify aria label on Custom Collection header/buttons to include the collection name

Cypress:
* Always rebuild image when running Cypress CI
* Added a `docker-compose.e2e.dev.yml` config for running services (Test IDP, Mongo, Redis) in order to debug Cypress while running the application locally
* Added some convenience Cypress commands:
    * `cy.preserveLoginCookies()` - preserves cookies required to maintain a user session between tests. Should be used in `beforeEach`
    * `cy.loginTestIDP()` - completes the log in flow with the test IDP and redirects back to the root URL
    * `cy.joinBeta()` - sets the beta cookie without having to explicitly visit `/joinbeta`

Fast follows:
- #292 
- #379 
- #404 

Fixes #126 

## Review Notes

To test, run the application stack locally in Docker and verify:
- You cannot access the application without logging in (all URLs should redirect you back to `/login`)
- You can login with both the test IDP and the production SAML environment (see instructions [here](https://github.com/USSF-ORBIT/ussf-portal-client/blob/main/docs/development.md#authentication-with-saml))
- Once logged in, you can access both the MVP and beta URLs. The beta pages should show your actual name in the header.
- You can log out from both the MVP and beta

image


abbyoung:
I have to click the Logout button twice for it to trigger -- is this expected, and/or just me?

e2e/cypress/integration/sites-and-applications.spec.js
```diff
@@ -2,9 +2,11 @@ describe('Sites and Applications', () => {
   before(() => {
     // Reset the database
     cy.task('db:seed')
+    cy.loginTestIDP()
   })
+
   beforeEach(() => {
-    // Make sure the beta cookie is set
+    cy.preserveLoginCookies()
     cy.visit('/joinbeta')
```
abbyoung, line 10: Is this where we'd use the `cy.joinBeta()` helper you mentioned in the PR notes, or is that for something else?

abbyoung, line 10: Is this where we'd use the `cy.joinBeta()` helper you mentioned in the PR notes, or is that for something else?

abbyoung (approved):
a mountain of work!! ✅ ✅ ✅ 

i'm only able to use test idp since my cac is 💀, but happy to report that flow works great.

jbecker01 (approved):
Looks good!

PR #402, opened 2021-12-03
ci: Attempting to automate manual deploys further

esacteksab:
## Description 

What I'm trying to do is based on this https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ that mirrors the C1 deployment leveraging a parameterized build. We don't need to pass a parameter because we use the git sha, but this should allow us to select a branch, and that's what we want, is to deploy a branch.


abbyoung (approved):

suzubara (approved):
Ooh nice!

PR #403, opened 2021-12-03
dev: Soft delete CMS-originated bookmarks in custom collections

### LINKED ISSUE #369: Modify delete CMS bookmark from collection to soft delete instead
Modify the Remove Bookmark functionality so that when a user removes a CMS bookmark from a collection, it uses a soft delete (ie, a "hidden" or "removed" flag) instead of hard deleting the record from the collection.
- Soft deleted bookmarks should be filtered out when rendering, so they don't appear in the UI
- If deleting a custom bookmark (that doesn't have a CMS ID) it can be hard-deleted

abbyoung:
## Description 

When a user deletes a bookmark from a custom collection in My Space, we want to soft delete it if it's a canonical bookmark from the cms.

The PR also creates a separate file to start holding Bookmark-specific queries instead of lumping them all in the resolvers.

* If the bookmark to be deleted is user-created, `Bookmark.deleteOne` will hard delete the item from the db.
* If the bookmark to be deleted is cms-created, `Bookmark.removeOne` will set `isRemoved` to true, and it won't be displayed.

Fixes #369

## Review Notes

To Do
- [x] Write and update tests
- [x] filter in query instead of ui
- [x] update seed data to include cmsId for canonical links


src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -242,13 +246,18 @@ const CustomCollection = ({
...
     <>
       <Collection header={customCollectionHeader} footer={addLinkForm}>
-        {bookmarks.map((bookmark: BookmarkType) => (
-          <RemovableBookmark
-            key={`bookmark_${bookmark._id}`}
-            bookmark={bookmark}
-            handleRemove={() => handleRemoveBookmark(`${bookmark._id}`)}
-          />
-        ))}
+        {bookmarks.map(
```
suzubara, line 249: minor nit: we should use `.filter().map()` since Array.map is meant to be used if each item is always going to return an item. ie:

```
{bookmarks.filter(b: BookmarkType => !b.isRemoved).map(...
```

src/components/MySpace/MySpace.test.tsx
```diff
@@ -246,20 +251,11 @@ describe('My Space Component', () => {
 
     userEvent.click(buttons[0])
 
-    // Wrapping this in act due to https://github.com/apollographql/apollo-client/issues/5920
     await act(async () => {
       jest.runAllTimers()
     })
 
-    jest.useRealTimers()
```
suzubara, line 254: I _think_ this should stay in?

src/components/MySpace/MySpace.test.tsx
```diff
@@ -209,33 +209,38 @@ describe('My Space Component', () => {
...
-            collectionId: mocks[0].result.data.collections[0].id,
+            _id: '3',
+            collectionId: '34',
+            cmsId: null,
           },
+          refetchQueries: [`getCollections`],
         },
         result: () => {
-          console.log('RESULT')
+          console.log('REMOVE_BOOKMARK RESULT')
```
suzubara, line 227: can we remove this?

src/components/MySpace/MySpace.test.tsx
abbyoung: i added it to `afterEach(() => ...` at the beginning of the file, so we wouldn't have to repeat in any other tests, but I can keep it here if you'd like!

src/components/CustomCollection/CustomCollection.tsx
abbyoung: thank you!

src/components/MySpace/MySpace.test.tsx
suzubara: ohh ok, in that case never mind!

suzubara (approved):

PR #419, opened 2021-12-09
feat: Expire session cookie

### LINKED ISSUE #292: Implement session timeout

suzubara:
## Description 

Sets the session cookie max age to 4 hours. After that time, a user will have to login again. This PR does not implement an idle timeout in the UI.

Fixes #292 

## Review Notes

Testing this as-is requires waiting 4 hours for your cookie to expire. If you run the app locally and are comfortable changing the code, you can modify this line in `src/lib/session.ts` to modify the expiration time:
```
-  maxAge: 60 * 60 * 4, // 4 hours
+  maxAge: 60, // 1 minute
```

Then the cookie should expire 1 minute after logging in.
- At this point you can refresh the page and verify you are redirected to the login page.
- Also test that the API errors: Log in to the beta My Space page, wait longer than a minute, and then try to perform an API action such as removing a link or adding a link. You should be redirected to the login page.


suzubara:
Hold on #408 so we can verify that making API calls will also error if the session has expired

abbyoung (dismissed):

abbyoung (approved):
works as expected! 👍

PR #453, opened 2021-12-16
feat: Custom error pages, Join beta button alignment

### LINKED ISSUE #431: Custom 404 page
We need to design & implement a 404 (page not found) page. The UX right now if a user goes to an unknown URL:

image

Implementation: https://nextjs.org/docs/advanced-features/custom-error-page#404-page

Also noted is that there is a bug when a user tries to directly accesses https://www.my.spaceforce.mil/sites-and-applications without having authenticated and without having the beta cookie set. The expected behavior is the user is routed back to the login screen. The actual behavior is that they are seeing the portal with a 404 error. 
image (8)

### LINKED ISSUE #432: Custom 500 error page
We need to design & implement a 500 (internal server error) page.

Implementation: https://nextjs.org/docs/advanced-features/custom-error-page#500-page

suzubara:
## Description 

This PR implements custom 404 & 500 pages in the portal app. It also performs an auth check on the 404 page and redirects the user to log in if they aren't (we may want to revisit this requirement when we redesign the Login page).

I also left-aligned the "Join beta" button on the MVP site per @MKDale 's request!

Fixes #431 
Fixes #432 

## Review Notes

You can view these pages locally by going to:
- http://localhost:3000/500
- http://localhost:3000/404

Or on the AWS environment:
- https://dev.ussforbit.us/404
- https://dev.ussforbit.us/500

image

image

image

### DIFF
--- a/public/assets/images/USSF_Vertical_Gradient_RGBicononly-01.svg
+++ b/public/assets/images/USSF_Vertical_Gradient_RGBicononly-01.svg
@@ -1,0 +1,159 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 25.4.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1"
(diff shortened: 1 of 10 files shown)


abbyoung (approved):

jbecker01 (approved):
Looked good on FF, Edge, Safari, and Chrome. Did not load properly on IE 11. Also, the bug this morning with being able to access /sites-and-applications w/o auth has been resolved.

esacteksab (approved):

PR #458, opened 2021-12-22
feat: Set up Matomo tracking on portal client

### LINKED ISSUE #316: Add Matomo tracking client to the portal pages
Add the Matomo analytics tracking code to the following pages:
- [x] Login 
- [x] My Space
- [x] All sites & applications

suzubara:
## Description 

This PR adds an analytics hook that initializes the Matomo tracker on each page, and fires page view events for SPA routing. If no Matomo URL or site ID are provided in the runtime env, it should silently fail. I included a `debug` flag so we can easily view analytics events in the browser console when desired.

It also adds Matomo and its database to the docker-compose services file for debugging purposes. The app should not fail if it tries to track events and Matomo is not available, or even if the Matomo env vars are undefined, so I don't know that we want to add Matomo to other compose files. We may also switch the Matomo image to build from our analytics repo in the future.

Fixes #316 

## Review Notes

To test:
- Run the app locally (`yarn install && yarn build && yarn start`)
- Run docker services (`docker compose -f docker-compose.services.yml up`)
- View Matomo locally: http://localhost/ (login is `User`, password `bitnami`)
- View app locally: http://localhost:3000

You should see your visits logging in the Matomo UI (check **Visitors > Real-time**). If you don't see any events, try viewing the portal in a browser other than Chrome.


next.config.js
```diff
@@ -14,6 +14,10 @@ const securityHeaders = [
 
 module.exports = withKeystone(
   withBundleAnalyzer({
+    publicRuntimeConfig: {
```
suzubara, line 17: @esacteksab this lets us define env vars for NextJS pages that can be defined at runtime (instead of build)

suzubara:
@esacteksab & I were unable to reproduce the missing env var issue so I removed the debugging code. This PR should be good to go!

esacteksab (approved):

abbyoung (approved):

PR #472, opened 2022-01-07
dev: split resolvers out and add unit tests

### LINKED ISSUE #371: Split resolvers into separate files and add unit tests
Move existing GraphQL resolver code that interacts with database into model files, and write tests

reference: https://www.apollographql.com/blog/backend/architecture/how-to-structure-graphql-server-code/

Also add limits for:
- Adding a link to a collection (no more than 10)
- Adding a collection to a user's my space (no more than 25)

abbyoung:
## Description 
* Refactors resolvers into operations per type: Bookmark, Collection
* Adds unit tests
* Adds max limit for Bookmarks (10)
* Adds max limit for Collections (25)
* Various bug fixes

Fixes #371


src/models/Bookmark.ts
```diff
@@ -64,9 +107,72 @@ export const BookmarkModel = {
       return {
         _id: _id,
       }
-    } catch (e) {
+    } catch (e: any) {
       console.error('Error in Bookmark.hideOne', e)
       return e
     }
   },
+  async addOne(
```
suzubara, line 115: this method makes me wonder if we should standardize on a cleaner function signature, for example group the object properties together and pass the context in separately. ie something like:

```
addOne({ collectionId, url, userId, label, cmsId }, { db, ...options })
```

suzubara (commented):
review is still WIP but left some comments in the meantime!

abbyoung:
i found a cool bug in my max bookmarks -- i need to filter out the hidden cms bookmarks when calculating how many bookmarks are in a collection

src/models/Bookmark.ts
abbyoung: updated -- take a look and let me know what you think! ⭐

suzubara: yes!! looks good :D I think it would be nice to roll this out to the other methods as well for the sake of consistency, but not a blocker for this PR

src/types/index.ts
```diff
@@ -57,6 +57,17 @@ export type BookmarkInput = {
...
+  _id: string
+}
+
+export type AddBookmarkInput = {
+  url: string
+  collectionId: string
+  userId: string
+  label?: string
+  cmsId?: string
+}
```
suzubara, line 70: For type definitions that are only relevant to a narrow scope of file(s) like these, I think it is fine to define them in the relevant file, and use `src/types` for more global types that are imported across different concerns. It's very easy to end up with a ton of type definitions for various things that get confused

src/models/Collection.ts
```diff
@@ -0,0 +1,198 @@
...
+        .collection('users')
+        .findOneAndUpdate(query, updateDocument, { returnDocument: 'after' })
+
+      // Return deleted collection id
+      return { _id: _id }
+    } catch (e) {
+      // TODO error logging
+      // eslint-disable-next-line no-console
+      console.error('error in remove collection', e)
+      return e
```
suzubara, line 195: I noticed a few different patterns for error handling (some places return, some throw, some swallow the error entirely) and I think we need to nail down a consistent pattern here, probably `throw` and ensure errors are surfaced back to the UI (so the user can be notified something went wrong). It would also be nice to establish conventions for the error messages since they'll be in the server logs, maybe something like: `CollectionModel Error: error removing collection`

suzubara (approved):
This is a big improvement over what we had! I have some comments that are mostly style/consistency-related but I don't want to block this from getting merged so make changes at your discretion and we can follow up later

src/models/Bookmark.ts
abbyoung: yeah happy to do that! wanted to make sure i was on track before doing all of them 😂

PR #482, opened 2022-01-14
fix: Fix bug for multiple bookmark removes 

### LINKED ISSUE #436: Unable to delete multiple URLs from a collection at a single time
To reproduce:
1) On My Space, click the "X" to remove a link from a collection
2) Immediately click the "X" on another link
3) Observe that when the Undo state ends, only the first link was removed

abbyoung:
## Description 
Fixes issue where removing two bookmarks consecutively would only remove the first.

Fixes #436 

## Review Notes
 Merge after #472


suzubara:
@abbyoung I decided to push the fix to your existing branch since you had added the Cypress test also! I believe this fixes the issue - I learned about `React.Children.map` which will iterate over `props.children` and generates its own unique key instead of using the index.

suzubara (approved):
@abbyoung says it works!

PR #487, opened 2022-01-20
feat: Add Link UX/UI improvements

suzubara:
## Description 

This PR makes the following enhancements to the Add Link flow:

* Custom links in a collection get a different UI treatment
* The Add Link ComboBox input only allows selecting an existing CMS link
* Adding a custom link is handled separately, in a modal
* Users can edit & delete custom links via modal

Also adds Storybook pages for all existing modal components.

Fixes #465 

## Review Notes

image

image

image

image

PR #492, opened 2022-01-26
feat: Edit collection title ux improvements

### LINKED ISSUE #464: The ability to edit the name of a collection comes when you click the name and is not available in the dropdown of the gear; this may not be obvious to people right away
Add an option to the dropdown menu from the cog icon to edit the collection name.

abbyoung:
## Description 

To Do
- [x] Review styles
- [x] Remove new collection if no title is entered / 'cancel' is clicked
- [x] Update unit tests

Fixes #464

### DIFF
--- a/e2e/cypress/database/seedMongo.js
+++ b/e2e/cypress/database/seedMongo.js
@@ -2,7 +2,7 @@
 const { MongoClient } = require('mongodb')
 const { testUser1, testUser2 } = require('./users.js')
 
-const DB = 'dev'
+const DB = 'cypress-e2e'
 
 async function dropCollection(mongoClient, collectionName) {
   const collection = mongoClient.db(DB).collection(collectionName)
--- a/e2e/cypress/database/seedMongo.js
+++ b/e2e/cypress/database/seedMongo.js
@@ -2,7 +2,7 @@
 const { MongoClient } = require('mongodb')
 const { testUser1, testUser2 } = require('./users.js')
 
-const DB = 'cypress-e2e'
+const DB = 'dev'
 
 async function dropCollection(mongoClient, collectionName) {
   const collection = mongoClient.db(DB).collection(collectionName)
--- a/src/components/CustomCollection/CustomCollection.tsx
+++ b/src/components/CustomCollection/CustomCollection.tsx
@@ -223,11 +223,17 @@ const CustomCollection = ({
     </Button>
   )
 
+  const [isActive, setActive] = useState(false)
+
   const editCustomCollectionTitle = (
     <Button
       type="button"
       className={styles.collectionSettingsDropdown}
-      onClick={handleEditCollectionTitle}>
+      onClick={() => {
+        console.log('clicked')
+        setActive(true)
+        console.log(isActive)
+      }}>
       Edit collection title
     </Button>
   )
@@ -247,6 +253,7 @@ const CustomCollection = ({
         text={title}
         onSave={handleEditCollection}
         onDelete={handleRemoveCollection}
+        isActive={isActive}
       />
 
       <DropdownMenu
--- a/src/components/CustomCollection/CustomCollection.tsx
+++ b/src/components/CustomCollection/CustomCollection.tsx
@@ -52,11 +52,14 @@ const CustomCollection = ({
     if (isAdding && linkInput.current) {
       linkInput.current.focus()
     }
+  }, [isAdding])
+
+  useEffect(() => {
     // If there is no title, prompt user to enter one
     if (title === '') {
       setEditing(true)
     }
-  }, [isAdding])
+  }, [])
 
   const handleSubmitEdit = (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault()
@@ -250,6 +253,13 @@ const CustomCollection = ({
     </>
   )
 
+  const handleCancelEdit = () => {
+    if (bookmarks.length === 0) {
+      handleRemoveCollection()
+    }
+    setEditing(false)
+  }
+
   const customCollectionHeader = (
     <>
       <EditableCollectionTitle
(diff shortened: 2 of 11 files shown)


src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -48,13 +46,29 @@ const CustomCollection = ({
   const urlInputValue = useRef<string>('')
   const addCustomLinkModal = useRef<ModalRef>(null)
   const linkInput = useRef<ComboBoxRef>(null)
+  const [isEditing, setEditing] = useState(false)
```
suzubara, line 49: maybe we should make these states a little more explicit now that there's a few things going on in this component? like, `isEditingTitle` and `isAddingLink`

src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -208,42 +222,71 @@ const CustomCollection = ({
...
+      <ol>
+        <li>{editCustomCollectionTitle}</li>
+        <li>{deleteCustomCollection}</li>
+      </ol>
     </>
   )
 
+  const handleCancelEdit = () => {
+    if (bookmarks.length === 0) {
+      handleRemoveCollection()
```
suzubara, line 258: this logic is confusing to me. Why do we want to remove the collection if there are no bookmarks and they cancel out of editing the title? I thought we are only removing the collection if they don't enter a value for the title (and there wasn't one previously set)

src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -48,13 +46,29 @@ const CustomCollection = ({
...
   }, [isAdding])
 
+  useEffect(() => {
+    // If there is no title, prompt user to enter one
+    if (title === '') {
+      setEditing(true)
+    }
+  }, [])
+
+  const handleSubmitEdit = (e: React.FormEvent<HTMLFormElement>) => {
```
suzubara, line 64: as a general rule, I would recommend keeping the event handler itself in the same component as the element that triggers it (in this case, the form) and just pass in the submission handler to that component. That way DOM API stuff is kept contained in the component that renders the elements.

src/components/CustomCollection/EditableCollectionTitle.tsx
```diff
@@ -1,91 +1,76 @@
...
+            required
             maxLength={200}
             className={styles.collectionTitle}
-            value={currentText}
-            placeholder="Add a title for this collection"
-            onKeyDown={(e) => handleKeyDown(e)}
-            onChange={(e) => setCurrentText(e.target.value)}
-            onBlur={handleOnBlur}
+            placeholder="Name this collection"
+            type="text"
```
suzubara, line 59: I would expect this input to have a default value of whatever the existing title is (and to reset to that value if the user cancels out). I had to implement a similar form in the Edit Custom Link modal if you want an example: https://github.com/USSF-ORBIT/ussf-portal-client/pull/487/files#diff-7557230bf066bb54052bf830361b2efbeacf03c1e953f31e360c71e8e80ec7aa

suzubara (changes requested):
I made a couple of suggestions! The main blockers to me are:
- The input should be pre-filled with the current title if editing an existing collection
- The logic around deleting the collection should just be if no title is entered, I don't think it should matter if there are bookmarks or not

src/components/CustomCollection/EditableCollectionTitle.test.tsx
```diff
@@ -15,231 +15,76 @@ describe('EditableCollectionTitle component', () => {
...
+    />
+  )
+  const input = screen.getByRole('textbox', { name: 'Collection Title' })
+  userEvent.type(input, 'New Title')
+
+  const saveNameButton = screen.getByRole('button', { name: 'Save name' })
+  userEvent.click(saveNameButton)
+  // #TODO "errors" when clicking submit on form; expected behavior by jsdom
+  // <<Error: Not implemented: HTMLFormElement.prototype.submit>>
+  // Should I ignore the error?
```
suzubara, line 71: I'm not seeing this warning, is this comment still applicable?

src/components/CustomCollection/EditableCollectionTitle.test.tsx
```diff
@@ -15,231 +15,76 @@ describe('EditableCollectionTitle component', () => {
...
+  expect(mockHandleOnSave).toBeCalledTimes(1)
```
suzubara, line 72: not a blocker but you can also use `toBeCalledWith('New Title')` to test that the handler was called with the expected param(s)

src/components/CustomCollection/EditableCollectionTitle.tsx
```diff
@@ -1,91 +1,86 @@
...
+  isEditing: boolean
 }
+
 export const EditableCollectionTitle = ({
   collectionId,
   text,
+  isEditing,
   onSave,
-  onDelete,
+  onCancel,
```
suzubara, line 27: looks like we can delete the `onDelete` prop from this component

src/components/CustomCollection/EditableCollectionTitle.tsx
```diff
@@ -1,91 +1,86 @@
...
 }: PropTypes) => {
-  const [isEditing, setEditing] = useState(text === '')
-  const [currentText, setCurrentText] = useState(text)
-  const inputRef = useRef<HTMLTextAreaElement>(null)
+  const inputRef = useRef<HTMLInputElement>(null)
 
+  const handleCancel = () => {
+    const inputEl = inputRef.current as HTMLInputElement
+    inputEl.value = ''
```
suzubara, line 33: since the `isEditing` flag changing causes the input element to re-render, I don't think we need to clear the value manually

src/components/MySpace/MySpace.tsx
```diff
@@ -53,7 +53,7 @@ const MySpace = ({ bookmarks }: { bookmarks: BookmarkRecords }) => {
             data.collections.map((collection) => (
               <Grid
                 key={`collection_${collection._id}`}
-                tablet={{ col: 6 }}
+                tabletLg={{ col: 6 }}
```
suzubara, line 56: so first, apologies, I realize the syntax for the grid classes is kind of confusing. This change is telling the layout to go from 1 item/row > 2 items/row at `tabletLg`, which I think we do want so the edit title form doesn't wrap. However I think we _also_ want to go from 2 items/row > 3 items/row at a larger size, so we also need:
```
- desktop={{ col: 4 }}
+ desktopLg={{ col: 4 }}
```

However both the `tabletLg, desktopLg` breakpoints are actually [turned off in USWDS by default](https://designsystem.digital.gov/utilities/layout-grid/) so these changes aren't having any effect. to fix, we need to turn them on in our `styles/_settings.scss` file by adding this:

```
$theme-utility-breakpoints: (
  'card': false,  // 160px
  'card-lg': false,  // 240px
  'mobile': false,  // 320px
  'mobile-lg': true,  // 480px
  'tablet': true,  // 640px
  'tablet-lg': true,  // 800px
  'desktop': true,  // 1024px
  'desktop-lg': true,  // 1200px
  'widescreen': false,  // 1400px
);
```

suzubara (commented):
this looks great! Everything works as expected. my comments are primarily around cleanup and also fixing the grid layout

src/components/MySpace/MySpace.tsx
abbyoung: thank you so much!!

e2e/cypress/database/seedMongo.js
```diff
@@ -2,7 +2,7 @@
 const { MongoClient } = require('mongodb')
 const { testUser1, testUser2 } = require('./users.js')
 
-const DB = 'cypress-e2e'
+const DB = 'dev'
```
abbyoung, line 5: whoops didn't mean to commit this -- need to revert

abbyoung:
@jbecker01 Deployed to dev and ready for your 👀 !

suzubara (approved):
💯

jbecker01 (approved):
looks good!

PR #495, opened 2022-01-27
feat: Add event tracking to existing features

### LINKED ISSUE #471: Add tracking events
We would also like to capture event actions on a page using trackEvent on the following buttons:

My Space page:
- [ ] Add link 
- [ ] Add section 
- [ ] Create new collection
- [ ] Select from template

All sites & applications:
- [ ] Add to My Space
- [ ] Example collections
- [ ] Add to new collection
- [ ] Sort alphabetically
- [ ] Sort by type
- [ ] Select multiple collections

suzubara:
## Description 

This PR converts the `useAnalytics` hook to a React context, to ensure the analytics script is only initialized once on the window, and the `push` method can be used across components without having to drill props down. I also added a `trackEvent` method that is just a shortcut for `push(['trackEvent'...`.

I added `trackEvent` to all user events in the [spec](https://docs.google.com/spreadsheets/d/1iQJOdfqQVk_WxmBmg0DoIPuc0EL_BwxtpT5MyoeJ8jE/edit#gid=0) that aren't being modified in other PRs.

Fixes #471 

## Review Notes

To test locally:
* Start up the app and Matomo using docker services
    *  `yarn build && yarn start`
    *  `docker-compose -f docker-compose.services.yml up -d`
* Log into the app and click through some of the actions that should fire events
* Verify that you see the events being tracked by Matomo 
    *  Matomo: http://localhost/
    * Log in: `User / bitnami`


src/stores/analyticsContext.tsx
```diff
@@ -50,6 +74,24 @@ const useAnalytics = ({
     windowWithAnalytics._paq.push(args)
   }
 
+  // Shortcut for push(['trackEvent', ...])
```
abbyoung, line 77: this is so great, I love how readable it makes the code wherever it's being used

abbyoung (approved):
💯 this is so good! 

do we need to add any documentation on when/where/how to use event tracking in the app? (not necessarily in this PR)

PR #525, opened 2022-02-23
dev(migration): Add type to Collections

suzubara:
## Description 

This adds a MongoDB migration to add the `type` field to existing items in `user.mySpace` with a value of `Collection` (in preparation for adding a second `user.mySpace` type).

Since this is our first MongoDB/DocDB migration it also adds some documentation and tooling to help test & run migrations. I refactored the server startup script to run outstanding migrations whenever the server starts, as well as to make it easier to add any other necessary scripts on startup.

Finally, this also includes code changes to:
- Write `type: 'Collection'` when creating new Collections so we don't need to run a backfill migration
- Include `type` in the `getCollections` query so that we can verify the new data in API responses

Fixes sc-301

## Review Notes

- Local: test the migration documentation locally & verify it works as expected
- AWS Dev: inspect `/graphql` response in browser dev tools on My Space and verify the collections data includes the `type: 'Collection'` field


.envrc
```diff
@@ -12,8 +12,9 @@
 # Environment Node is running in (including build scripts)
 # NOT the same as the environment the application is running in
 # Use "development" for local development with hot reloading, otherwise use "production"
-# Options: development, production
-export NODE_ENV=development
+# Options: development, production, test
```
esacteksab, line 15: If we're just going to comment this out, why not get rid of it? Isn't that the reason we use source control?

esacteksab (dismissed):
Had to do some Googling, but looks right. 😀

.envrc
suzubara: Yes true.. I thought the comments were worth keeping though because `NODE_ENV` can be a confusing thing sometimes

esacteksab (dismissed):

esacteksab (approved):

esacteksab (approved):

abbyoung (dismissed):

esacteksab (approved):

abbyoung (approved):

PR #530, opened 2022-02-25
dev: Update Docker Compose files for local dev

abbyoung:
## Description 

* Clean up Docker Compose files and makes them DRY
* Adds yarn scripts to control bringing docker containers up/down for services

Fixes #sc-277

### Testing local dev
* Run `yarn services:up` - starts all services required 
* Or run `yarn cms:up` - starts all services required plus external Keystone
* ...then run `yarn dev` to start Portal

### Testing Cypress locally
* Run `yarn e2e:up` to start services including Cypress-specific MongoDB instance
* Run `yarn dev` to start app
* Run `yarn cypress:dev` to run tests


docker-compose.cms.yml
```diff
@@ -0,0 +1,20 @@
+services:  
+  cms:
+    container_name: keystone-cms
+    # image: keystone-cms
```
abbyoung, line 4: need 2 remove

docker-compose.services.yml
```diff
@@ -74,9 +74,23 @@ services:
       - 'matomo_data:/bitnami/matomo'
     depends_on:
       - mariadb
+# CMS-specific services      
+  keystone-db:
+    container_name: keystone-db
+    image: postgres:latest
```
esacteksab, line 80: Can we please pin this rather than `latest`?

abbyoung: my b - thanks for catching! is pinning to `postgres:14` sufficient?

esacteksab: Yes. It may be that we find ourselves here with a newer version but that wouldn't necessarily be a bad thing. PostgreSQL is often very good with forward/backward compatibility and I don't think we're doing anything too exotic.

docker-compose.cms.yml
```diff
@@ -0,0 +1,19 @@
...
+    container_name: keystone-cms
+    build:
+      context: https://github.com/USSF-ORBIT/ussf-portal-cms.git#main
+      dockerfile: Dockerfile
+    restart: always
+    ports:
+      - '3001:3001'
+    environment:
+      - PORT=3001
+      - SESSION_SECRET='bHkJ6Js~=E~1(yw@a!hB`zFgA.QcaHbHkJ6Js~=E~1(yw@a!hB`zFgA.QcaH'
```
suzubara, line 12: I know it doesn't matter what this value is and that it has to be set to something but it makes me feel a little uneasy that it _looks_ like it might matter so what if we set it to something like `SESSION_SECRET='thisvaluecanbeanythingitisonlyforlocaldevelopment'`

docker-compose.cms.yml
```diff
@@ -0,0 +1,19 @@
...
+      - DATABASE_URL=postgres://keystone:keystonecms@keystone-db:5432/keystone
+      - TEST_USERNAME='Admin'
+      - TEST_PASSWORD='adminpassword'
+      - TEST_EMAIL='admin@example.com'
```
suzubara, line 16: I don't think these values should be quoted, I had to enter the value _with_ quotes in order to login.

suzubara (approved):
looks good to me!! My only comment was a minor hiccup :) thanks for documenting!

suzubara (approved):

PR #535, opened 2022-02-28
chore(deps): Update dependencies

suzubara:
## Description 

Updates dependencies minor/patch versions that require no code changes, and removes resolutions we don't need to manage anymore (because the direct dependency updated)


esacteksab (dismissed):
<3

esacteksab (approved):

PR #552, opened 2022-03-07
dev: Add tracing

abbyoung:
# SC-64

## Proposed changes

* Adds `tracing.js` which initializes open telemetry/xray tracing.
* Implements auto-instrumentation using the node and web metapackages (ALL THE THINGS)

---

## Setup

* When branch is deployed to dev, you should be able to see _lots_ of logs in X-Ray tracing.

---

---


startup/tracing.js
```diff
@@ -0,0 +1,51 @@
...
+// OTEL tracing
+const { BatchSpanProcessor } = require('@opentelemetry/tracing')
+const { NodeTracerProvider } = require('@opentelemetry/node')
+const { OTLPTraceExporter } = require('@opentelemetry/exporter-trace-otlp-grpc')
+const { AWSXRayPropagator } = require('@opentelemetry/propagator-aws-xray')
+const { AWSXRayIdGenerator } = require('@opentelemetry/id-generator-aws-xray')
+const { trace } = require('@opentelemetry/api')
+
+const { detectResources } = require('@opentelemetry/resources')
+const { awsEc2Detector } = require('@opentelemetry/resource-detector-aws')
```
esacteksab, line 11: Should this be `awsECSDetector` instead? Also, line 16 below?

package.json
```diff
@@ -43,6 +43,17 @@
...
+    "@opentelemetry/auto-instrumentations-node": "^0.27.4",
+    "@opentelemetry/exporter-trace-otlp-grpc": "^0.27.0",
+    "@opentelemetry/id-generator-aws-xray": "^1.0.1",
+    "@opentelemetry/instrumentation-aws-sdk": "^0.5.1",
+    "@opentelemetry/node": "^0.24.0",
+    "@opentelemetry/propagator-aws-xray": "^1.0.1",
+    "@opentelemetry/resource-detector-aws": "^1.0.3",
+    "@opentelemetry/resources": "^1.0.1",
+    "@opentelemetry/sdk-trace-node": "^1.0.1",
+    "@opentelemetry/tracing": "^0.24.0",
```
suzubara, line 56: can we pin all of the above pls?

src/pages/api/sysinfo.ts
```diff
@@ -1,16 +1,55 @@
...
+    return JSON.stringify({ traceId: xrayTraceId })
+  }
+  return
+}
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  console.log('Responding to /sysinfo')
+  const tracer = await tracing.initTracing()
```
suzubara, line 32: does this need to be init again if it happened at startup?

package.json
abbyoung: YES this was just from testing -- thank you for the reminder!

src/pages/api/sysinfo.ts
abbyoung: all this code was deleted since it was just for testing the hookup cause no, it doesn't

suzubara (approved):

PR #570, opened 2022-03-21
refactor: convert _id from type string to objectid

abbyoung:
# SC-303

## Proposed changes

* Convert all `_id` types from string to ObjectId
* `_id` and `collectionId` represent a MongoDB id, and will always be ObjectId

### Tasks
- [x] Confirm jest tests all passing
- [x] Confirm cypress tests all passing  
- [x] Clean up bson vs mongodb module and usage of objectId constructor
- [x] Meet minimum test coverage for jest (90%)
  

## Reviewer notes

~Need to review some cases with @suzubara where strings are implicitly required (GraphQL related?). Also discuss when to use bson vs mongodb ObjectId module and if it's okay to mix + match.~

~Notes for Suz:~

~* Unit tests of concern that are currently not passing are in `resolvers/index.test.ts`.~
 ~* There are still some GraphQL errors of what it expects/what it's getting with the ObjectId and custom scalar.~
~* The issue I was running into using bson's `new ObjectId()` in some of the fixtures was it caused matching problems in the test. Instead of the id itself, it would print out `new ObjectId(0123456789etc)`. So the usage of 'mongodb' module and its ObjectId is still hanging around some test files.~
~* Still not sure where to find `Resolvers` type that supported passing in a custom scalar~

## Setup

Make sure all the tests pass, and you can still use the app as expected.

---

- [x] Created/modified automated unit tests in Jest
  - [x] Including jest-axe checks when UI changes
- [x] Created/modified automated E2E tests in Cypress
  - [x] Including cypress-axe checks when UI changes


esacteksab (approved):
I pulled the branch and ran it locally. Everything appears to work as I'd expect. Admittedly, I don't quite understand why the change, but that's a me problem. 😝

src/models/Bookmark.test.ts
```diff
@@ -1,16 +1,22 @@
-import { ObjectId } from 'bson'
-import { MongoClient, Db } from 'mongodb'
+import { MongoClient, Db, ObjectId } from 'mongodb'
+import { ObjectId as typeObjectId } from 'bson'
```
suzubara, line 2: nit: can we call this `ObjectIdType` ?

src/models/Bookmark.ts
```diff
@@ -84,6 +86,13 @@ export const BookmarkModel = {
     { db }: Context
   ) {
     try {
+      console.log('BookmarkModel: editOne', {
```
suzubara, line 89: Can we remove these stray console logs?

suzubara (approved):
this is a ⛰️ of work and it looks awesome! Especially nice increase in tests :) My comments are all tiny nits and not necessarily blockers but might be easier to go ahead with now..

src/models/Bookmark.ts
abbyoung: bless u ✨

src/models/Bookmark.test.ts
abbyoung: yes!

src/components/BookmarkList/BookmarkList.tsx
```diff
@@ -5,7 +5,7 @@ import {
   IconExpandLess,
   IconExpandMore,
 } from '@trussworks/react-uswds'
-
+import { ObjectId } from 'bson'
```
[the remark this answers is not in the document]
abbyoung: yes thank you!

suzubara (approved):

PR #571, opened 2022-03-22
chore: Update dependencies

suzubara:
# Dependency updates (3/22/2022)

## Proposed changes

This PR makes minor/patch dependency updates that don't require code changes. There are some resulting formatting (prettier) changes.

---

## Reviewer notes

All tests should pass.

---


src/pages/api/bookmarks.ts
```diff
@@ -1,5 +1,6 @@
 import type { NextApiRequest, NextApiResponse } from 'next'
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
suzubara, line 3: `eslint-plugin-import` is complaining about these imports not being defined in package.json but they are not external imports. This will become a moot point when embedded Keystone is removed.

esacteksab (approved):
THANK YOU!!!

PR #601, opened 2022-04-12
chore: Test fetching content from Keystone

abbyoung:
# SC-292

## Proposed changes

* Adds test page at `/keystone-test` to display content from external CMS
* Adds `ApolloLink` in `apolloClient.tsx` to handle splitting requests based on context
  * If the request should hit Keystone, the mutation/query must have `clientName: 'cms'` passed in to the context
* Adds env var exposed to the browser with Keystone base url
 
**To Do**
- [x] Update Docker compose files with env vars
## Setup
~*NOTE*: Testing this PR is technically blocked by [chore: Migrate existing cms data to postgres](https://github.com/USSF-ORBIT/ussf-portal-cms/pull/31) in ussf-portal-cms. If you've pulled this branch and ran the migration locally, you _should_ be able to continue with testing if you do not reset your docker volume.~

It also requires the cms server config in the corresponding PR https://github.com/USSF-ORBIT/ussf-portal-cms/pull/32

* Pull down ussf-portal-cms branch `sc-292-connect-to-portal` and run locally
* Pull down this branch and run locally
* Visit `/keystone-test` and confirm you can see collections and bookmarks displayed

  - ... 
  - [ ] Checked that all Storybook accessibility checks are passing
  - [ ] Including jest-axe checks when UI changes
  - [ ] Including cypress-axe checks when UI changes
- Performed [a11y testing](https://github.com/trussworks/accessibility/blob/master/sample_a11y_testing_process.md):
  - [ ] Checked responsiveness in mobile, tablet, and desktop
  - [ ] Checked keyboard navigability
  - [ ] Tested with [VoiceOver](https://dequeuniversity.com/screenreaders/voiceover-keyboard-shortcuts) in Safari
  - [ ] Checked VO's [rotor menu](https://github.com/trussworks/accessibility/blob/master/README.md#how-to-use-the-rotor-menu) for landmarks, page heading structure and links
  - [ ] Used a browser a11y tool to check for issues (WAVE, axe, ANDI or Accessibility addon tab for Storybook)
- [ ] Requested a design review for user-facing changes
- For any new migrations/schema changes:
  - [ ] Followed guidelines for zero-downtime deploys

- Performed [a11y testing](https://github.com/trussworks/accessibility/blob/master/sample_a11y_testing_process.md):
  - [ ] Checked responsiveness in mobile, tablet, and desktop
  - [ ] Checked keyboard navigability
  - [ ] Tested with [VoiceOver](https://dequeuniversity.com/screenreaders/voiceover-keyboard-shortcuts) in Safari
  - [ ] Checked VO's [rotor menu](https://github.com/trussworks/accessibility/blob/master/README.md#how-to-use-the-rotor-menu) for landmarks, page heading structure and links
  - [ ] Used a browser a11y tool to check for issues (WAVE, axe, ANDI or Accessibility addon tab for Storybook)

### As a designer reviewer, I have

- [ ] Checked in the design translated visually
- [ ] Checked behavior
- [ ] Checked different states (empty, one, some, error)
- [ ] Checked for landmarks, page heading structure, and links
- [ ] Tried to break the intended flow
- Performed [a11y testing](https://github.com/trussworks/accessibility/blob/master/sample_a11y_testing_process.md):
  - [ ] Checked responsiveness in mobile, tablet, and desktop
  - [ ] Checked keyboard navigability
  - [ ] Tested with [VoiceOver](https://dequeuniversity.com/screenreaders/voiceover-keyboard-shortcuts) in Safari
  - [ ] Checked VO's [rotor menu](https://github.com/trussworks/accessibility/blob/master/README.md#how-to-use-the-rotor-menu) for landmarks, page heading structure and links
  - [ ] Used a browser a11y tool to check for issues (WAVE, axe, ANDI or Accessibility addon tab for Storybook)

### As a test user, I have

- Run through the [Test Script](hhttps://docs.google.com/spreadsheets/d/1eV8UK0aJZ0qrzjnXf5SJ_uRiV7bpsj3yrhRFEkjOvV4/edit?usp=sharing):
  - [ ] On commercial internet in IE11
  - [ ] On commercial internet in Edge
  - [x] On commercial internet in Firefox
  - [x] On commercial internet in Chrome
  - [x] On commercial internet in Safari
  - [ ] On NIPR in IE11
  - [ ] On NIPR in Firefox
  - [ ] On NIPR in Chrome
  - [ ] On NIPR in Safari
  - [ ] On a mobile device in Firefox
  - [ ] On a mobile device in Chrome
  - [ ] On a mobile device in Safari
- [ ] Added any anomalous behavior to this PR

---


abbyoung:
Related to https://github.com/USSF-ORBIT/ussf-portal-cms/pull/32

jcbcapps (dismissed):

src/pages/keystone-test.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+  const collections = cmsCollections?.collections as CollectionRecords
+
+  const {
+    loading: cmsBookmarksLoading,
+    error: cmsBookmarksError,
+    data: cmsBookmarks,
+  } = useQuery(GET_KEYSTONE_BOOKMARKS, {
+    context: {
+      clientName: 'cms',
```
abbyoung, line 78: @jcbcapps this is the code that tells the app which API to hit -- we'll need to pass in this context with `clientName: 'cms'` with any query or mutation that we're sending to keystone.

For internal api calls, we don't need to pass in the context (it'll default to the portal api).

src/pages/keystone-test.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+  const handleAddSelected = () => {
+    const collectionObjs = selectedCollections.map((id) =>
+      collections.find((i) => i.id === id)
+    ) as CollectionRecords
+
+    const collectionTitles = collectionObjs.map((c) => c.title).join(',')
+    trackEvent('S&A add collection', 'Add selected', collectionTitles)
+
+    handleAddCollections({
```
suzubara, line 143: Trying to add collections from this page gives me a GraphQL error, it looks like the Keystone API is returning `__typename` which this mutation is not expecting as part of the object.

suzubara (commented):
I know that this is a temporary test page and what we should be able to do with it was a little fuzzy, but I think it will be useful to test the migration from the sqlite data to Keystone data. To that end, maybe it would be worth essentially duplicating the sites and applications E2E test but testing against this page instead?

src/pages/keystone-test.tsx
jcbcapps: I was actually just looking into this on my branch. Should the change be made on this branch, or could I just update the CollectionRecord type to include the `__typename` field?

.envrc
```diff
@@ -35,6 +35,13 @@ export SAML_ISSUER=ussf-portal-client
 export SESSION_SECRET=thisvaluecanbeanythingitisonlyforlocaldevelopment
 export SESSION_DOMAIN=localhost
 
+
+##############################
+# Keystone Configuration     #
+##############################
+export NEXT_PUBLIC_KEYSTONE_URL=http://localhost:3001
```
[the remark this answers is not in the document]
suzubara: If we are actually only using the Keystone URL from the server, does this still need to be a `NEXT_PUBLIC...` var? My understanding of those is that they get baked in if they are defined at build time and can be a little confusing, so if we aren't using this in any browser code then it might be best to rename to just `KEYSTONE_URL`.

src/pages/api/sysinfo.ts
```diff
@@ -12,5 +13,6 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
     nodeEnv: __NODE_ENV__,
     analyticsUrl: process.env.MATOMO_URL,
     analyticsSiteId: process.env.MATOMO_SITE_ID,
+    keystoneUrl: __KEYSTONE_URL__,
```
suzubara, line 16: I noticed that I don't actually see any value here if the `NEXT_PUBLIC_KEYSTONE_URL` var is not defined until after building the image, and I think this needs to be treated like the Matomo vars:
```
keystoneUrl: process.env.NEXT_PUBLIC_KEYSTONE_URL
```

the pattern of injecting this value into the `next.config.js` file and using `delcare const` here is only for variables whose value is known at _build_ time (it's a webpack thing)

next.config.js
```diff
@@ -23,6 +23,9 @@ module.exports = withKeystone(
           __VERSION__: JSON.stringify(process.env.npm_package_version),
           __BUILD_ID__: JSON.stringify(buildId),
           __NODE_ENV__: JSON.stringify(process.env.NODE_ENV),
+          __KEYSTONE_URL__: JSON.stringify(
```
suzubara, line 26: I think we can remove this here, see my other comment

src/operations/mutations/addCollections.tsx
```diff
@@ -1,16 +1,30 @@
...
 interface AddCollectionsResponse {
   collections: Collection[]
 }
 
 interface AddCollectionsInput {
-  collections: CollectionRecords
+  collections: CollectionRecordInput[]
+}
+
+export const addCollectionsInput = (c: CollectionRecordInput[]) => {
```
suzubara, line 17: Ok, my understanding of this function is that it's taking Collection data from the CMS (`CollectionRecords`), and mapping it to the input expected by the `addCollections` operation to be inserted into Mongo (`CollectionRecordInput[]`), right? I was a little bit confused & I think this is a good place where we could use more explicit types to make the purpose of the function a little clearer.

For example, if these are the type definitions:
```
export type BookmarkRecord = {
  id: string
  url: string
  label?: string
  description?: string
}

export type CollectionRecord = {
  id: string
  title: string
  bookmarks: BookmarkRecords
}

export type CollectionRecordInput = Pick<
  CollectionRecord,
  'id' | 'title' | 'bookmarks'
>

export type BookmarkRecordInput = Pick<BookmarkRecord, 'id' | 'url' | 'label'>
```

the function can be typed like so:
```
export const addCollectionsInput = (
  c: CollectionRecords
): CollectionRecordInput[] => {
  return c.map(({ id, title, bookmarks }) => {
    return {
      id,
      title,
      bookmarks: bookmarks.map(({ id, url, label }): BookmarkRecordInput => {
        return { id, url, label }
      }),
    }
  })
}

``` 

Also I don't think this function needs to add the `type` property, because that is explicitly added by the `Collection.addMany` model.

src/pages/keystone-test.tsx
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  const isSelected = (id: string): boolean =>
+    selectedCollections.indexOf(id) > -1
+
+  const handleAddSelected = () => {
+    // Find selected collections in the array of all CMS collections
+    const collectionObjs = selectedCollections.map((id) =>
+      collections.find((i) => i.id === id)
+    ) as CollectionRecordInput[]
```
suzubara, line 125: I think if you make the type changes suggested, you won't need to cast this using `as CollectionRecordInput` - because at this point, this data is not actually of that type, and is actually `CollectionRecords`

src/pages/sites-and-applications.tsx
```diff
@@ -118,14 +123,14 @@ const SitesAndApplications = ({
   const handleAddSelected = () => {
     const collectionObjs = selectedCollections.map((id) =>
       collections.find((i) => i.id === id)
-    ) as CollectionRecords
+    ) as CollectionRecordInput[]
```
suzubara, line 126: same comment as above, I think we can avoid typing this altogether

suzubara (commented):
I tested & the main functionality works as expected! I left some comments suggesting changes around the environment variable & type definitions

.envrc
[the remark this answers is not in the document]
abbyoung: will we only ever use keystone on the server?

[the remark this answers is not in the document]
suzubara: we probably will end up referencing it in both server & browser. I'm really just suspicious of `NEXT_PUBLIC_` because that was how I had tried to define other env vars (like the Matomo URL) and there were issues when deploying 🤷‍♀️

suzubara (approved):
👏 👩‍🍳 :shipit:

suzubara (approved):

PR #618, opened 2022-04-19
ci: Check for required env vars on startup

suzubara:
# SC-475

## Proposed changes

Adds a check to the startup script that throws an error if any required environment variables are undefined. Also updates documentation around setting/adding env vars.

## Reviewer notes

To test, edit your local environment to remove some required env vars and try starting the app in dev or production mode:
```
yarn dev
```
```
yarn build && yarn start
```

Instead of starting the app, it should throw an error that includes the name of the env var that is missing.

---

### DIFF
--- a/src/lib/mongodb.ts
+++ b/src/lib/mongodb.ts
@@ -14,22 +14,11 @@ declare global {
   }
 }
 
-const host = process.env.MONGO_HOST || ''
-const user = process.env.MONGO_USER || ''
-const password = process.env.MONGO_PASSWORD || ''
-
-const RDS_TLS_CERT = process.env.RDS_TLS_CERT || 'rds-combined-ca-bundle.pem'
-
 // Connection string for DocumentDB
-const connectionString =
-  process.env.MONGO_URL ||
-  `mongodb://${user}:${password}@${host}/?tls=true&tlsCAFile=${RDS_TLS_CERT}&replicaSet=rs0&readPreference=secondaryPreferred&retryWrites=false`
+const connectionString = process.env.MONGO_URL
 
 let client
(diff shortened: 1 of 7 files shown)


src/lib/mongodb.ts
```diff
@@ -14,22 +14,11 @@ declare global {
...
-const user = process.env.MONGO_USER || ''
-const password = process.env.MONGO_PASSWORD || ''
-
-const RDS_TLS_CERT = process.env.RDS_TLS_CERT || 'rds-combined-ca-bundle.pem'
-
 // Connection string for DocumentDB
-const connectionString =
-  process.env.MONGO_URL ||
-  `mongodb://${user}:${password}@${host}/?tls=true&tlsCAFile=${RDS_TLS_CERT}&replicaSet=rs0&readPreference=secondaryPreferred&retryWrites=false`
+const connectionString = process.env.MONGO_URL
```
suzubara, line 18: I cleaned this up because as far as I can tell, `MONGO_URL` is defined in all environments (included local dev) and so the connection string logic is no longer used. @abbyoung do you mind confirming this looks right to you?

abbyoung: looks good to me!

esacteksab (approved):

abbyoung (approved):
works as expected! thanks for adding the documentation -- i'll update my other PR with a new env var once this is merged 😄

jcbcapps (approved):

jcbcapps (approved):

abbyoung (approved):

src/lib/mongodb.ts
suzubara: reader: it was not

PR #635, opened 2022-05-12
chore: Dependency updates

suzubara:
- Updates dependencies that don't require code changes. Remaining dependency updates have explicit stories for handling.
- Removed packages & resolutions we are no longer using


esacteksab (approved):
❤️

PR #648, opened 2022-05-19
refactor: Sync all canonical url updates

jcbcapps:
# SC-516

## Proposed changes

- Retrieve bookmark URLs and labels from CMS instead of Mongo to make sure they are kept up to date

## Reviewer notes

Log in as `cmsadmin / cmsadminpass` and:
- In the Keystone UI, edit the URL (or label) of an existing bookmark that is in one of the user's collections.
- Navigate to the user's My Space, refresh the page and verify that you see the change that was made in Keystone.

## Setup
Be sure to have the latest version of the CMS on your local machine and running.

---

### DIFF
--- a/src/components/CustomCollection/CustomCollection.test.tsx
+++ b/src/components/CustomCollection/CustomCollection.test.tsx
@@ -172,6 +172,24 @@ const mockLinks = [
     url: 'http://www.example.com/3',
     label: 'Test Bookmark 3',
   },
+  {
+    id: 'cmsId3',
+    label: 'vMPF',
+    url: 'https://afpcsecure.us.af.mil/',
+    __typename: 'Bookmark',
+  },
+  {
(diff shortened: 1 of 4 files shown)


abbyoung:
Works great! So happy to have this synced up.

I checked out the data in mongo to see if/how it changed, and we do have the label and url for these canonical urls stored.

Should we change how we're storing this data in mongo? If we never want to use the label or url from the portal db, we don't need them. I'm not positive if that would break existing code / if we should store null values there instead of omitting the fields entirely. This might be a bigger discussion for a different PR.

I believe it'd include updating mocks and test data, and new user creation with the example collection. Not sure it needs to be in the PR but wanted to flag it!

Screen Shot 2022-05-20 at 8 58 39 AM

abbyoung (approved):

PR #664, opened 2022-06-15
chore: adding Tall John to codeowners

esacteksab:
# SC-601

Just like the title says.


.github/CODEOWNERS
```diff
@@ -1,11 +1,15 @@
 # Default owners for everything
-* @suzubara @abbyoung @esacteksab
+
+- @suzubara @abbyoung @esacteksab @gidjin
```
suzubara, line 3: I don't understand why `*` was changed to `-` here... maybe an overzealous auto formatter?

esacteksab: Yes indeed. Thanks for catching this.

PR #715, opened 2022-07-21
chore: snapshotting docdb and deleting docdb snapshots

esacteksab:
This updates the script to snapshot docdb to actually wait for the completion of the snapshot process.

This also includes documentation on how to manually delete docdb snapshots. This ensures that in my absence, someone can delete these should the build get blocked because of 100 manual snapshots existing. 

Also noticed that `deploy-prototype` script lingering around, but this environment never saw the light of day, no longer exists, so removing the script.


scripts/snapshot-docdb.sh
```diff
@@ -1,6 +1,23 @@
...
+
+# Don't do anything else until it's available
+# aws docdb does not have a `wait`
+until [ "$DBSS" = "available" ]
+    do
+      DBSS=$(aws docdb describe-db-cluster-snapshots --db-cluster-snapshot-identifier $SSID |jq -r '.DBClusterSnapshots[]' | jq -r '.Status')
+      echo $DBSS
+    done
+    echo "$SSID available!"
+    continue
```
jcbcapps, line 23: Bash scripts aren't my strong suit so forgive the dumb question, but do you need a new line at the end of the file?

esacteksab: I'm not sure it matters. Do you have a linter going off?

esacteksab: I installed shellcheck and it doesn't complain about the lack of a new line.

jcbcapps: No I'm not seeing any errors, I was just curious. Looking in the GitHub UI it has that little red circle and I just thought that was a warning or something.

esacteksab: I hadn't noticed that. Good catch.

gidjin (approved):
LGTM

PR #43, opened 2021-07-22
dev: content list item component

abbyoung:
## Description 

WIP / not ready for review

This PR splits out the Content List as two components: Item and Group. For use on the home page rebuild.

* Adds `ContentListGroup` component and tests
* Adds `ContentListItem` component and tests

Fixes (part of) #16 

## Review Notes
`yarn storybook` to view the component group with one item

Issue #10, opened 2021-07-02
Add COVID banner component


suzubara:
Create a component for rendering the COVID alert: https://github.com/USSF-ORBIT/ussf-portal-mvp/blob/main/_includes/banner-alert.html

PR #716, opened 2022-07-21
chore(deps): addresses #10 and #43 dependabot alerts

esacteksab:
Tests pass locally and should fix [this](https://github.com/USSF-ORBIT/ussf-portal-client/security/dependabot/10) and [this](https://github.com/USSF-ORBIT/ussf-portal-client/security/dependabot/43).


gidjin (approved):
LGTM

PR #732, opened 2022-08-04
ci: filter paths for github actions

esacteksab:
As discussed in eng sync and similar to https://github.com/USSF-ORBIT/ussf-portal-cms/pull/107 this filters paths so github actions don't always run unnecessarily.


gidjin (approved):
Looks good to me.

Question: What about `storybook-branch.yml`, and `man-dev-*` workflows? Do those need this exclusion as well? If not I'm good with this going through. If they should then feel free to add them and merge it.

PR #765, opened 2022-09-01
ci: build docker image plus cache and push to ecr

minhlai:
# SC-916

## Proposed changes

- build docker image with cache in-line
- push image with cache to AWS ECR, allowing other workflows to utilize the cache to speed up builds

---

## Reviewer notes

Once merged to Main, the workflow can be run on dispatch and tested there. There is no anticipated effect on any existing workflows.

## Setup

---

### DIFF
--- a/.github/workflows/cache-to-ecr.yml
+++ b/.github/workflows/cache-to-ecr.yml
@@ -7,11 +7,6 @@ on:
       - 'docs/**'
       - '**.md**'
     branches: [main]
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '**.md**'
-    branches: [main]
 
 jobs:
   build-and-push-dev-aws:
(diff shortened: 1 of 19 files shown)


.github/workflows/cache-to-ecr.yml
```diff
@@ -0,0 +1,77 @@
...
+on:
+  workflow_dispatch:
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - '**.md**'
+    branches: [main]
+
+jobs:
+  build-and-push-dev-aws:
```
gidjin, line 12: Question: A lot of this looks like what's in `.github/workflows/dev-aws-ecr.yml`. Should we at some point reuse the common steps? If so I don't think it needs to be part of this story.

gidjin (approved):
LGTM, assuming we can get the e2e test to actually pass. I believe we need USSF-ORBIT/ussf-portal#144 to be merged first, though.

.github/workflows/cache-to-ecr.yml
minhlai: Yes I think at some point it makes sense for this to be a reusable workflow so we do not repeat code.

PR #772, opened 2022-09-09
feat: update user model for saving dark mode

gidjin:
# SC-1030

## Proposed changes

Add `theme` as an attribute of the `User` model. 

## Reviewer notes

I refactored the `User.editOne` to be `User.setDisplayName` as well as added tests for its matching resolver. I did modify it to return data so that the test in the resolver could be written the same way as the others. Please let me know if this breaks something or violates a convention.

I declared `theme` to be of type `'light' | 'dark'` so that auto complete and typescript would know that it should only ever be `light` or `dark`, at least until we add `system` or other themes.

## Setup

I'm not sure there is a way to try this out yet since there is no graphql setup yet. But sc-1035 is next on my list so that should make it testable.

For now ensure tests pass
```sh
yarn test
```

---

- [X] Met the acceptance criteria, or will meet them in subsequent PRs or stories
  - SC-1035 will add a mutation to GraphQL to set theme, and ensure it's returned as part of the find user calls
  - SC-1031 will ensure that the saved setting is used when logging in
- [x] Created/modified automated unit tests in Jest
- For any new migrations/schema changes:
  - [x] Followed guidelines for zero-downtime deploys

  - [ ] Checked that the E2E test build is not failing

### DIFF
--- a/src/models/Bookmark.test.ts
+++ b/src/models/Bookmark.test.ts
@@ -33,7 +33,10 @@ describe('Bookmark Model', () => {
     // Create test user
     const testUserId = 'testUserId'
     const displayName = 'Floyd King'
-    await User.createOne(testUserId, [exampleCollection], displayName, { db })
+    const theme = 'light'
+    await User.createOne(testUserId, [exampleCollection], displayName, theme, {
+      db,
+    })
     const testUser = await User.findOne(testUserId, { db })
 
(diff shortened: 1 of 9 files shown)


abbyoung (approved):
looks great!

jcbcapps (approved):

PR #774, opened 2022-09-13
feat: add mutation for get and set theme

gidjin:
# SC-1035

## Proposed changes

This work is dependent on PR #772 and story sc-1030 and should be merged AFTER them. It adds the query to return the theme for the current user and the mutation to update the theme. Also modifies the apollo context to set a default theme if one is not present.

Since I was in the file that configures the Apollo server I also [applied their recommendation to bound the cache](https://www.apollographql.com/docs/apollo-server/performance/cache-backends/). As far as I can tell it is already using an internal cache but it is unbounded. Eventually we may want to configure it to use redis. This is for sc-707

## Reviewer notes

This is my first time modifying graphql stuff, please let me know if I missed something or didn't follow a convention I am not aware of.

## Setup

See the steps in `docs/development.md` to enable the Apollo Server Explorer. Then execute a query to get theme or set theme.

```graphql
query Query {
  displayName
  theme
}
```

```graphql
mutation EditTheme {
  editTheme(userId: "FLOYD.KING.376144527@testusers.cce.af.mil", theme: "dark") {
    theme
  }
}
```
---

- [X] Met the acceptance criteria, or will meet them in subsequent PRs or stories
  - SC-1030 will update the model to support a theme PR #772 
  - SC-1031 will ensure that the saved setting is used when logging in
  - [ ] Checked that the E2E test build is not failing

  - [ ] Checked that the E2E test build is not failing

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -7,6 +7,7 @@
 1. [Database migrations](#database-migrations)
 1. [PR linting](#pr-linting)
 1. [Testing](#testing)
+1. [Apollo Server Explorer](#apollo-server-explorer)
 1. [Releasing](#releasing)
 
 ## Environment Setup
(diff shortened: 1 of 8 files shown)


src/pages/api/graphql.tsx
```diff
@@ -92,25 +93,37 @@ export const apolloServer = new ApolloServer({
...
+          await User.createOne(userId, [initCollection], displayName, 'light', {
+            db,
+          })
         } catch (e) {
           // TODO log error
           // console.error('error in creating new user', e)
           throw new ApolloError('Error creating new user')
         }
+      } else {
+        // TODO we should be able to remove these once all exisitng users have the defaults set.
```
abbyoung, line 108: question/thought for later: how will be know when this happens?

docs/development.md
```diff
@@ -7,6 +7,7 @@
 1. [Database migrations](#database-migrations)
 1. [PR linting](#pr-linting)
 1. [Testing](#testing)
+1. [Apollo Server Explorer](#apollo-server-explorer)
```
abbyoung, line 10: 🥳 yay documentation!!!

abbyoung (approved):
excellent! thanks to your docs was able to easily test query and mutation, and validate in mongodb. 🎉

src/pages/api/graphql.tsx
jcbcapps: I guess we would just have to do this periodically, but maybe we could write a db query looking for any user that doesn't have the defaults set? Then at some point in the future, the query won't return anything and we'll know? There's probably an easier way, but that's just the first thought I had.

gidjin: yeah I think we'd have to run queries against the data, or think of an approach that was like a db migration where on deploy it looped through all our users and added the defaults if they are missing and would only be run once. Things are simple enough I'm okay if we just manually run a query at some point in the future and once it shows everyone has the defaults we can drop the code. 

There might be other ways though and if either of you dig into it or want to lets get it on our list for tech debt.

gidjin: I added a story in shortcut about this discussion sc-1128

PR #782, opened 2022-09-22
fix: remove unnecessary trailing double quote

minhlai:
## Proposed changes

- remove a harmless but unnecessary trailing double quote from Dockerfile
---


gidjin (approved):
LGTM

PR #789, opened 2022-09-28
chore: upgrade to 14.20.1

esacteksab:
I think I got all the places. This upgrades us to Node 14.20.1 which addresses a few CVE's.


minhlai (approved):
LGTM!

PR #844, opened 2022-11-02
fix: design: dark mode improvements

shkeating:
# SC-1276

first attempt at dark mode audit findings 

## Proposed changes

- sets padding at bottom of `main` element instead of margin to prevent gap when gradient is applied
- adds background gradient to dark mode
- sets up color variables to ensure there is a color fallback for gradient background
- sets breadcrumbs to use correct color
- sets correct background color on myspace component
- adds dark background on breadcrumbs page in storybook, breadcrumb links are always on a dark background

## Reviewer notes

[figma page ref](https://www.figma.com/file/iPOdDkRxCaeoDoQ4JXjxlk/Space-Force-Service-Portal?node-id=9415%3A60627)

## Setup

   
    yarn storybook
    

---

## Screenshots

localhost_3000_

localhost_3000_ (1)


src/components/BreadcrumbNav/BreadcrumbNav.stories.tsx
```diff
@@ -8,7 +8,7 @@ export default {
   component: BreadcrumbNav,
   decorators: [
     (Story) => (
-      <div className="sfds">
+      <div className="sfds darkBgLock">
```
gidjin, line 11: Question: Is this something we only need in storybook?

gidjin (approved):
Looks good to me, I left one minor question. Also this will need design approval to merge :)

src/components/BreadcrumbNav/BreadcrumbNav.stories.tsx
shkeating: I'd say yes. when this component is used anywhere in the app, it always has a dark background

gidjin: Sounds like we should apply the `darkBgLock` class to the `BreadCrumbBar` component in our `BreadcrumbNav` instead of only adding it here. Maybe this is a good topic for Design Eng pairing  session?

jcbcapps (approved):
Looks great!

PR #852, opened 2022-11-03
chore: resolve apollo-server-core@3.11.1

minhlai:
This deals with [dependabot alert 54](https://github.com/USSF-ORBIT/ussf-portal-client/security/dependabot/54) and [Trivy scan result #89](https://github.com/USSF-ORBIT/ussf-portal-client/security/code-scanning/89)

#845 did not resolve the issue because apollo-server-micro was causing apollo-server-core to resolve to an earlier version. Both packages will be updated to 3.11.1.


minhlai:
This PR was closed because the branch name was inaccurate.

PR #853, opened 2022-11-03
chore: resolve apollo-server-core@3.11.1

minhlai:
This deals with [dependabot alert 54](https://github.com/USSF-ORBIT/ussf-portal-client/security/dependabot/54) and [Trivy scan result #89](https://github.com/USSF-ORBIT/ussf-portal-client/security/code-scanning/89)

#845 did not resolve the issue because apollo-server-micro was causing apollo-server-core to resolve to an earlier version. Both packages will be updated to 3.11.1.

Changelog: https://github.com/apollographql/apollo-server/blob/version-3/CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,8 +55,8 @@
     "@opentelemetry/sdk-trace-node": "1.3.1",
     "@opentelemetry/tracing": "0.24.0",
     "@trussworks/react-uswds": "3.2.0",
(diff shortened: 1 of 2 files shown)


gidjin (approved):
LGTM, everything appears to be working.

PR #861, opened 2022-11-16
fix: design: few more dark mode improvements

shkeating:
# SC-1276

## Proposed changes
- switches Edit Name to be green in both dark and light mode
- sets View More News button to use tertiary button styling and removes styles trying to overwrite default button styles

## Setup

```
yarn storybook
yarn dev
```

## Screenshots
image
 im having trouble running the app locally but the button is now the right colors in both modes.


gidjin (approved):
LGTM, fyi if you prefix your branch with `sc-1276` (eg `sc-1276-dark-mode-two`) next time github will automatically associate your branch with the shortcut story. 

I'm approving but having Mallory take a look might be good too.

malworks (approved):
Looks good!

PR #867, opened 2022-11-21
chore: replace deprecated set-output in gha

esacteksab:
# SC-1371

This replaces the deprecated `set-output` with the new format per the [docs](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)


minhlai (approved):
LGTM

PR #877, opened 2022-11-30
test: Mock rss url in storybook

gidjin:
## Proposed changes

Happo regularly flags the RSS widgets for changes since they pull live data, so everytime an article is published Happo flags it as a change. This PR installs `storybook-addon-mock` to support mocking the call and still allow us to see what live data would look like via the control panel. This should make it so happo will not flag it when new articles are posted after this is merged.

## Reviewer Notes

Are there other components that need this that I can add before merging?

Example of happo run I'm trying to fix https://happo.io/a/909/p/1321/compare/594701348446048ece35639813e838d993e3a16b/b866bc50cc3c299f18f0bf91a70a3dbefed84083

## Setup

* Run `yarn install` to install the new dep
* Start story book `yarn storybook`
* View `My Space > Example My Space` and `Sections > News > Space Force RSS` in storybook
* If you want to see live data
  * click the `Mock Request` panel
  * click `On` for the request
  * click the `Remount component` button next to doc at the top of the page (looks like refresh button)   

## Screenshots
SCR-20221130-ken
SCR-20221130-kew


package.json
```diff
@@ -150,6 +150,7 @@
     "sass-loader": "13.1.0",
     "standard-version": "9.5.0",
     "storybook-addon-apollo-client": "4.0.13",
+    "storybook-addon-mock": "^3.2.0",
```
jcbcapps, line 153: Do we need to lock this version in?

jcbcapps (approved):
Looks great! Left one non-blocking question. Also, the `ArticleList` component can accept items from the RSS feed but we already have some hard-coded examples in the storybook component. Thought it worth mentioning if you want to refactor it to use the mock request.

package.json
gidjin: Good catch, we should since we lock everything down. I'll fix it.

gidjin:
> Looks great! Left one non-blocking question. Also, the `ArticleList` component can accept items from the RSS feed but we already have some hard-coded examples in the storybook component. Thought it worth mentioning if you want to refactor it to use the mock request.

Maybe we can pair on this, I was initially only mocking the ones that called the `useRSSFeed` hook, I'm not seeing where that is for `ArticleList`.

jcbcapps:
@gidjin I misspoke. It's `news-announcements.tsx` that uses the `useRSSFeed` hook and passes each news item to `ArticleList` as a prop. But since `ArticleList` doesn't actually use the hook itself, I don't think it's anything to worry about. Sorry for the confusion 🙂

shkeating (approved):

abbyoung (approved):

PR #883, opened 2022-12-07
chore: design: storybook: Reorganization of Storybook Components

shkeating:
# SC-1423

## Proposed changes

This PR reorganizes the categories of our Storybook library to make it more easily navigable. This is the result of an [evaluation and re-sorting ](https://www.figma.com/file/SlNLNQhHEEtrYqAro9vVOq/Storybook-Organziation?node-id=1%3A167&t=xEpmXLRIgDZDG4Gq-1) conducted recently by design

Previously, we had three main storybook sections: **Global**, **Layout**, and **Components**. This PR expands on these previous categories like so:
- **Global** remains the same in principle, that it shows base styles and things we see central to the USSF brand. This PR removes **Button** from this category and moves **Logo** in.
- **Layout** was renamed to **Navigation** and adds in the **PageNav** component.
- **Components** was split into 3 new categories: 
> - **Base** where simple components without a corresponding parent component reside
> - **Components** where more complex, custom components go. If they have a corresponding child component (such as **ArticleList** and **ArticleListItem**, they get nested into that component using a folder.
> - **Layouts**, where our most complex components go. They often combine several components into an even more complex one, and these components may also be used for other things, too.

This branch renames two components:
- **BookmarkList** is renamed to **ApplicationTable**, as the **BookmarkList** had no relationship to the **Bookmark** component, which was confusing.
- **NewsListItem** is renamed to **NewsItem** because the name implied it was the child of the **NewsList** component that dies not exist.

This branch creates two stories pages:
- **SearchResults** to follow the pattern of the other components with -Item in their name. This nests **SearchResultItem** and ensures it gets the same level of importance to other comparable components.
- **Welcome** which is meant to be an informative landing page. I intend to drop my notes about how this is organized there so that when new components are added, there is guidance on the best place to put them.

## Reviewer notes
- look through what I have on the Welcome page, does it make sense? anything I should add?
- This will cause a lot of happo changes because it will change how many component file names are saved within happo
- I do not currently have access to my local dev environment for the actual app. Please take a peek or pair with me so we can check to enure this did not affect any app code. the only css change made was adding space below search result items.

## Setup

```
yarn storybook
```

## Screenshots
Storybook Organziation

Screen Shot 2022-12-08 at 17 32 53


gidjin (approved):
LGTM! I like the new organization

malworks (approved):
Looks great!

PR #886, opened 2022-12-09
fix: design: updates page loader animation

shkeating:
# SC-1434

## Proposed changes

This branch updates the page loader animation to look a bit cleaner and use imagery more associated with what USSF does. I used some public domain vectors of an earth and satellite instead of a moon and rocket, because it felt more in line with the work of USSF.

## Reviewer notes
I can't see how this looks in the app so please check that adding a background fill doesn't mess anything up

## Setup

    ```
    yarn storybook
    ```
http://localhost:6006/?path=/story/components-loader--default-loader 
---

## Screenshots

loader


src/components/Loader/Loader.tsx
```diff
@@ -8,19 +8,10 @@ const Loader = () => (
...
     <div className={styles.container}>
       <div className={styles.planet}>
-        <div className={styles.circle}></div>
-        <div className={styles.circle}></div>
-        <div className={styles.circle}></div>
-        <div className={styles.circle}></div>
-        <div className={styles.circle}></div>
-        <div className={styles.circle}></div>
-        <div className={styles.circle}></div>
+        <img src="/assets/images/earth.svg" alt=" " />
```
gidjin, line 11: Question: should the alt tag here be something like `alt="earth"` or `alt="cartoon representation of earth from space"`? Does that type of thing matter for a loader?

gidjin (approved):
Looks good and works locally in the app too. I left a question but it's not blocking.

src/components/Loader/Loader.tsx
shkeating: the loader itself has a screenreader label that serves in place of the whole thing

shkeating: I made it empty on purpose because we specifcally *don't* want a screenreader to output something like this

> loading.... Image: earth image: sattelite

we'd just want it to say `loading...` to them

jcbcapps (approved):
Looks great!

PR #895, opened 2022-12-21
fix: design: adds icons into theme toggle

shkeating:
# SC-1383

## Proposed changes

adds font awesome icons into theme toggle to have sun and moon show next to theme names.

## Reviewer notes

I was not able to match the implementation of font awesome icons on the other components I referenced, and imported them the way I knew how. Please check that this way is okay. I don't mind refectoring to get it in alginment with how they rest of the components import icon assets.

Please check that the dark mode button looks right in production. The storybook view does not switch over to dark mode after clicking the button, not even on staging.

## Setup

    
    yarn storybook

go to **Theme Toggle**

---

## Screenshots
image


malworks (approved):
I can't speak to the code but it LOOKS right! 😛

PR #899, opened 2023-01-05
test: fix the ThemeToggle storybook so it toggles

gidjin:
## Proposed changes

When reviewing #895 I noticed that the `ThemeToggle` button didn't switch themes in storybook. This PR fixes that.

## Reviewer notes

Clicking the `ThemeToggle` will switch the theme, if you then click the storybook theme toggle it won't work the first time but will on subsequent ones. I believe that's because the state for that one is controlled differently but it seemed super minor to me.

## Setup

```sh
yarn storybook
```

Visit http://localhost:6006/?path=/story/base-themetoggle--with-user

Marvel at the toggling. See the screenshots included below for examples.

## Screenshots
SCR-20230105-ei9

SCR-20230105-ehv


shkeating (approved):
im waiting for happo to come thru, but it's looking good for me locally! awesome.

PR #904, opened 2023-01-10
chore: update storybook to v6.5.15

minhlai:
## Proposed changes
- This update pgrades loader-utils to 2.0.4 in storysource and source-loader, getting us almost all the way there with replacing vulnerable loader-utils versions

---


esacteksab (approved):

shkeating (approved):
LGTM doesnt look like it messed woth our happo output, which would be my only real

PR #923, opened 2023-02-03
feat: Display video in an article

jcbcapps:
# SC-1541

## Proposed changes
Embeds a video into a CMS article

Note: connected to [this PR](https://github.com/USSF-ORBIT/ussf-portal-cms/pull/260), which adds the component block to the article schema

---

## Reviewer notes
One thing I've noticed is some sort of CORS issue (see screenshot below). Seems that YouTube is trying to set a cookie and our CORS policy is blocking it? 

## Setup
Checkout `jc-investigate-video` branch in the CMS and create an article w/ an embedded video
- Instructions [here](https://github.com/USSF-ORBIT/ussf-portal-cms/pull/260)

---

## Screenshots

cors-issue


src/components/SingleArticle/SingleArticle.tsx
```diff
@@ -28,6 +30,28 @@ export const SingleArticle = ({ article }: { article: ArticleRecord }) => {
   })
   const publishedDateObj = new Date(publishedDate)
 
+  const componentBlockRenderers: InferRenderersForComponentBlocks<
+    typeof componentBlocks
+  > = {
+    embedVideo: (props: any) => {
+      // We are receiving a YouTube link from the CMS, so we need to get the last section of the url.
+      // If there is a link, we split the url by '/' and get the last part to add to our embed link below.
+      const youtubeLink = props.link ? props.link.split('/').pop() : null
```
gidjin, line 39: Suggestion: Seems like we could trust the url sent by the CMS and use it as is without having to recreate it. For example `src={props.link}` instead of the below?

src/components/SingleArticle/SingleArticle.tsx
```diff
@@ -28,6 +30,28 @@ export const SingleArticle = ({ article }: { article: ArticleRecord }) => {
   })
   const publishedDateObj = new Date(publishedDate)
 
+  const componentBlockRenderers: InferRenderersForComponentBlocks<
```
gidjin, line 33: Question: is it possible to add a unit test for this function?

gidjin (changes requested):
Left one suggestion that I think should be implemented. One question that isn't blocking.

src/components/SingleArticle/SingleArticle.tsx
jcbcapps: Good catch. I must've been tired when I wrote this. I need to move this to the CMS repo, then we can accept the prop here as is like you said.

src/components/SingleArticle/SingleArticle.tsx
jcbcapps: Not sure? I'll dig into that and see what I come up with

src/components/SingleArticle/SingleArticle.tsx
jcbcapps: Wanted to give an update on this. I'm not entirely sure how to go about accomplishing this. I attempted to move this little snippet of code to the CMS repo where the link is being added, and then use a hook with `resolveInput` to format the link the way we want it, but it seems like we can't use hooks on a component block? Idk, I'm still playing around with it. I feel like it's possible, but haven't figured it out yet.

gidjin: Maybe we should pair, I'm not following why we need the snippet of code at all vs just using the url but feels like I'm missing some nuance.

shkeating (commented):
are we tracking this in storybook anywhere currently?

gidjin (approved):

shkeating (approved):
lookin good!

PR #953, opened 2023-03-08
ci: only push to c1 artifactory on release publication

minhlai:
## Proposed changes

- We decided to push to C1 Artifactory only when a release is published rather than every time something is pushed to Main. A manual workflow dispatch can still be used to push any branch to Artifactory.
---


gidjin (approved):
LGTM

PR #964, opened 2023-03-24
fix: enable matomo to track local dev portal

minhlai:
# SC-1720

## Proposed changes
- provides docker compose environment variables needed for matomo
- remove https port from `test-saml-idp` (unused) which conflicts with matomo https port
- add matomo and mariadb as services in `docker-compose.services.yml` so matomo starts for these commands: `yarn serivces:up`, `yarn cms:up`, and `yarn portal:up`
- 

---

## Setup

1. And enter the command `direnv allow` in portal-client repo so the new environment variables are applied.
2. Make sure you have [USSF-ORBIT/analytics repo has been cloned and is in the parent folder of the current portal-client repo.
3. Make sure your browser is not set to send `Do Not Track` requests to websites and you do not have a adblocker or script blocker running.

---

## Code review steps
There are several ways to start Matomo and Portal Client.
- In `USSF-ORBIT/analytics` enter the command `docker compose up -d` to start Matomo

You can start Portal Client in a couple ways:
- Start CMS and Portal Services in `ussf-portal-client` with `yarn cms:up` and then start Portal with `yarn dev`
- Start everything in a container with `yarn portal:up`

Log into Matomo at https://localhost:8443 with credentials in 1Pass.


gidjin (commented):
Question: Can we update the port in the `.envrc` instead of having everyone change `.envrc.local`? Also were you able to see events in Matomo? If so we should pair. If not might be good to ensure events are making it before calling this done.

Another question: should we start matomo with `docker-compose.services.yml` like we do the database? Good discussion maybe for sync or eng sync.

abbyoung (approved):
tested locally, lgtm!

PR #967, opened 2023-03-28
fix: design: storybook: link color adjustments and storybook documentation

shkeating:
# SC-1560
# SC-1644

## Proposed changes

This branch adjusts the dark background base link styles and ensures it is [displayed correctly in storybook](http://localhost:6006/?path=/story/global-typography--links). 

This branch also adjusts link colors that were catching color contrast violations on a few components including:
- the [logout button](http://localhost:6006/?path=/story/navigation-header--default-header) (this also aligns it to what we have in figma) -> text changed from white to black
- the [news item tag ](http://localhost:6006/?path=/story/components-newswidget-newsitem--rss-article)(i know its design is going to change but in the meantime i aligned it more to what its changing to) -> text color darkened and background lightened
- the 'Here's how you know' link in the gov banner. this actually turned out to be compliant but we thought it was just hard to read so why not improve it. 

## Setup

```
yarn storybook
```
http://localhost:6006/?path=/story/global-typography--links
http://localhost:6006/?path=/story/components-newswidget-newsitem--rss-article
http://localhost:6006/?path=/story/navigation-header--default-header

## Screenshots

Screen Shot 2023-03-28 at 14 24 11

Screen Shot 2023-03-28 at 14 24 07

image

image

image

### DIFF
--- a/src/components/NewsItem/NewsItem.module.scss
+++ b/src/components/NewsItem/NewsItem.module.scss
@@ -92,6 +92,7 @@ a.articleLink {
 
 .articleTag {
   align-self: flex-start;
+  color: $black;
 }
 
 .newsWidgetItem {
--- a/src/components/NewsItem/NewsItem.tsx
+++ b/src/components/NewsItem/NewsItem.tsx
@@ -38,7 +38,9 @@ const NewsItem = ({
         </span>
       </p>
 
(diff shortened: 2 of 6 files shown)


jcbcapps (approved):
Looks great!

abbyoung (approved):
lgtm! 🎉

malworks (approved):

PR #973, opened 2023-03-30
fix: design: adds background fill settings box 

shkeating:
# SC-1807

## Proposed changes

Adds background fill css to the widget box container surrounding the box on the settings page

sets text fields in light mode to default to white to account for layouts like this, they were transparent before, and we dont want text fields to become illegible if they are on an unexpected background in the future.

## Setup

```
yarn dev
```

http://localhost:3000/settings

---

## Screenshots
Screen Shot 2023-03-30 at 15 21 26

Screen Shot 2023-03-30 at 15 21 32

### DIFF
--- a/src/styles/pages/settings.module.scss
+++ b/src/styles/pages/settings.module.scss
@@ -10,6 +10,7 @@
 
   .widgetContainer {
     border: 1px solid $light-gray;
+    background: var(--background-fill);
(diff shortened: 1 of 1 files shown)


gidjin (approved):
LGTM, but if you want to wait for @malworks to approve that's fine with me.

PR #975, opened 2023-03-31
fix: Remove any null bookmarks and prevent future ones

gidjin:
# SC-1814

## Proposed changes

Somehow a user has a collection with a bookmark of value `null`. This PR changes our graphql schema to require bookmark objects in the collections objects to prevent this from happening again. It also introduces a migration that removes any null bookmark records from all users.

## Reviewer notes

Collection schema was changed so try as many collection actions as you can. If you want to test the migration use mongo express to add null as a record in bookmark then run the migration

To clear db of nulls
```
yarn migrate up 1680214667334-remove-null-bookmarks.js
```

To reset migration - does nothing to the data
```sh
yarn migrate down 1680214667334-remove-null-bookmarks.js
```

## Setup

Standard startup 

```sh
yarn services:up
cd ../cms && yarn dev
cd ../client && yarn dev
```

### DIFF
--- a/migrations/1680214667334-remove-null-bookmarks.js
+++ b/migrations/1680214667334-remove-null-bookmarks.js
@@ -1,0 +1,23 @@
+'use strict'
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { runMigration } = require('../utils/mongodb')
+
+module.exports.up = runMigration(async (db) => {
+  await db.collection('users').updateMany(
+    {},
+    {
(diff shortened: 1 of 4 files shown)


abbyoung:
To fix the linter error with graphql changes, we need to change `editCollection.graphql` to have the updated required `BookmarkReorderInput!`. Then we should be good to go!

abbyoung (approved):
lgtm 🚢

PR #985, opened 2023-04-10
feat: fully release docs page and search

gidjin:
## Proposed changes

This PR removes LD flags for documentation page and search box so they will be available for everyone.

## Reviewer notes

These are on in dev, staging, and production, so nothing should look different.

## Setup

Standard

```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```


abbyoung (approved):
LGTM!There's one small lint error for an unused var that could be corrected, but that's it.

shkeating (approved):

PR #996, opened 2023-04-14
build: download ca bundle without strict cert check and verify sha256 sum

minhlai:
## Proposed changes

- added a checksum file for Certs in DoD PKI CA Cert bundle from GH Secrets
- modfiy `add-dod-cas.sh` to download Cert bundle without strict TLS verification
- modify `add-dod.cas.sh` to verify the sha256 checksum of the certs that are downloaded against the checksum file to deter MITM attacks
---

## Code review steps

The workflow `.github/workflows/build-push-c1-art.yml` was tested with a manual dispatch and the results can be seen here: https://github.com/USSF-ORBIT/ussf-portal-client/actions/runs/4724289278

The output of the `Add DoD Certifaction Bundle` should include this:
```
 Verification successful
certificates_pkcs7_v5_11_dod_der.p7b: OK certificates_pkcs7_v5_11_dod_dod_root_ca_3_der.p7b: OK certificates_pkcs7_v5_11_dod_dod_root_ca_4_der.p7b: OK certificates_pkcs7_v5_11_dod_dod_root_ca_5_der.p7b: OK certificates_pkcs7_v5_11_dod_dod_root_ca_6_der.p7b: OK certificates_pkcs7_v5_11_dod_pem.p7b: OK dod_pke_chain.pem: OK README.txt: OK
```
Which indicates that the checksum has passed.

### DIFF
--- a/scripts/add-dod-cas.sh
+++ b/scripts/add-dod-cas.sh
@@ -14,8 +14,6 @@
     unzip unclass-certificates_pkcs7_DoD.zip
 
     # check that Checksums verify
-    # output=$(cd certificates_pkcs7_v5_11_dod; openssl smime -verify -in ../certificates_pkcs7_v5_11_dod.sha256 -inform DER -CAfile dod_pke_chain.pem | dos2unix | sha256sum -c)
-    # echo 123 >> certificates_pkcs7_v5_11_dod/certificates_pkcs7_v5_11_dod_dod_root_ca_5_der.p7b
     output=$(cd certificates_pkcs7_v5_11_dod; sha256sum -c --strict ../dod_ca_cert_bundle.sha256)
     echo $output
     if [[ "$output" == *"FAILED"* ]]; then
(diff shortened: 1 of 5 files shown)


scripts/add-dod-cas.sh
```diff
@@ -4,11 +4,20 @@
 
     # Location of bundle from DISA site
     bundle=https://dl.dod.cyber.mil/wp-content/uploads/pki-pke/zip/unclass-certificates_pkcs7_DoD.zip
-
+    # Copy the Checksum file
+    cp ./scripts/certificates_pkcs7_v5_11_dod.sha256 /usr/local/share/ca-certificates/
     # Extract the bundle
     cd /usr/local/share/ca-certificates
-    wget $bundle
+    wget --no-check-certificate $bundle
```
gidjin, line 11: Thought: when the certificate issue is resolved might be nice to just leave the sha check in anyway even if we can remove the `--no-check-certificate` flag.

gidjin (approved):
LGTM

PR #1017, opened 2023-05-01
refactor: Investigate using a new drag-and-drop library

jcbcapps:
# SC-1347

## Proposed changes
- Rough implementation of [dnd-kit](https://github.com/clauderic/dnd-kit) to replace our current drag-and-drop library

---

## Reviewer notes
This is a basic implementation of another dnd library just to show that it's possible given our current set up.

A few important notes:
- ~CUSTOM BOOKMARKS DON'T WORK. If you have any in your local db, delete them if you want to test this branch. The reason for this is that I just haven't added them back in yet because I think it will be a little tricky, and as mentioned I've already spent a fair amount of time on this.~ I fixed this. Custom bookmarks should be fine now.
- When adding a bookmark, you need to refresh the page to see it
- When deleting a bookmark, you need to refresh as well
- Drag-and-drop works, but doesn't actually perform a mutation to update the DB
- There are failing tests that I haven't even looked at 
- I set `reactStrictMode` back to true

I figured that I have spent enough time on this investigation, and before spending anymore time on a full implementation, I'd show off what I have so far to see what others think.

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---


src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -295,38 +324,71 @@ const CustomCollection = ({
...
+            const sortedBookmarkCopy = { ...sortedBookmark }
+            const visibleBookmarksCopy = [...visibleBookmarks]
+
+            visibleBookmarksCopy.splice(sortedBookmarkIndex, 1)
+            visibleBookmarksCopy.splice(i, 0, sortedBookmarkCopy)
+            setVisibleBookmarks(visibleBookmarksCopy)
+          }
+        }
+        return sortedItems
+      })
```
jcbcapps, line 367: I hope the comment I left makes sense. But if not, here is a short summary: dnd-kit has an extension that allows for sorting of a list of draggable items. In order to use that sorting functionality, we need to pass an array of unique ids (unfortunately, it has to be an array of strings or something similar, can't just use the bookmark objects), for which I used the _id of each bookmark. When you drop a draggable item, the previously mentioned array is updated in state to represent the new order of the items. We also need to update the order of `visibleBookmarks` because it contains all of the actual bookmark details, and we need to use it in our mutation to update what is in the db.

next.config.js
```diff
@@ -25,9 +25,7 @@ module.exports = withBundleAnalyzer({
 
     return config
   },
-  // TODO: This change was made so that drag and drop would continue to work after upgrading to React 18.
-  // Change this back to true once we find another way to keep drag and drop working
-  reactStrictMode: false,
+  reactStrictMode: true,
```
gidjin, line 28: 🎉

src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -58,6 +70,21 @@ const CustomCollection = ({
 }: PropTypes) => {
   const linkInput = useRef<ComboBoxRef>(null)
 
+  const [, setActiveDragId] = useState(null)
```
gidjin, line 73: Question: How does this get used? Does something else check the state variable outside of this file?

gidjin (commented):
Looks good, lots of changes. I think this would be a good thing to walk through in eng sync this week.

src/components/CustomCollection/CustomCollection.tsx
jcbcapps: Oh I forgot to remove this. dnd-kit has a component named `<DragOverlay />` that basically just lets you define some jsx and additional styles while dragging an item. So I can make the bookmark look completely different if I wanted to while dragging it. The `activeId` was just acting as a flag to display the `<DragOverlay />` with its children when a drag event starts.

PR #1020, opened 2023-05-10
build: remove unused saml certs

gidjin:
## Proposed changes

We use the SAML provider metadata to retrieve the current certs for verification. This means we do not need to stash a copy in our build.

## Reviewer notes

- Are there other references to these?
- [ ] We should deploy this to dev before merging to ensure everything still functions as expected.

## Setup

Start up is a little different so that everything is in a docker container.

### Start the system
```sh
yarn services:removeall
docker compose build
yarn portal:up
```


minhlai:
There is also this [script](https://github.com/USSF-ORBIT/ussf-portal-client/blob/5d98779ae388dbf06c5e9c6de7b66758861b71fa/scripts/create-gcds-chain.sh) that gets run in the [Dockerfile](https://github.com/USSF-ORBIT/ussf-portal-client/blob/5d98779ae388dbf06c5e9c6de7b66758861b71fa/Dockerfile#L83) which combines several certificates including the SAML cert. We should probably remove the `/usr/local/share/ca-certificates/federation.*.cce.af.mil.crt` certs from this script for this test.

```shell
# Make Dev GCDS PEM
for cert in "/usr/local/share/ca-certificates/DoD_Root_CA_3.crt" "/usr/local/share/ca-certificates/DOD_EMAIL_CA-59.crt" "/usr/local/share/ca-certificates/federation.dev.cce.af.mil.crt"; do
    (cat "${cert}") >> Dev-GCDS.pem;
done
```

I'm not sure if the first two certs are needed inside the image either. I know we need at least the first one in the workflow in order to connect to Artifactory and that's reflected in the[ .github/workflows/build-push-c1-art.yml](https://github.com/USSF-ORBIT/ussf-portal-client/blob/5d98779ae388dbf06c5e9c6de7b66758861b71fa/.github/workflows/build-push-c1-art.yml#L14) workflow.


And then there's this Environment Variable in the [Dockerfile](https://github.com/USSF-ORBIT/ussf-portal-client/blob/5d98779ae388dbf06c5e9c6de7b66758861b71fa/Dockerfile#LL108C1-L108C68), which points to a file that never gets created. Instead. this env var is overridden in AWS by the ECS Container Definition Environment Variable which points to `Dev-GCDS.pem`, `Test-GCDS.pem`, or `Prod-GCDS.pem` according to which environment.

```shell
ENV NODE_EXTRA_CA_CERTS='/usr/local/share/ca-certificates/GCDS.pem'
```

I would recommend removing this line.

gidjin (commented):
LGTM @minhlai. Probably should have @abbyoung or @jcbcapps look at this PR since we both worked on it

minhlai:
> LGTM @minhlai. Probably should have @abbyoung or @jcbcapps look at this PR since we both worked on it

Sounds good. I can explain it in a future PR review session too.

minhlai (approved):
We went over this in Eng Sync.

PR #1030, opened 2023-05-30
fix: a11y: adds error suggestion to url field in custom link modal

shkeating:
# SC-2023

## Proposed changes
changes error message in add a custom link modal to include guidance around expected input when a user enters an invalid url

## Setup

   
    yarn dev
    

- log in http://localhost:3000/
- initiate process to add a custom link to a collection
- enter an invalid url
- see new error messaging

---

## Screenshots

image


minhlai (approved):
LGTM! Went through the steps to add an invalid custom link and got the updated error message. Wonder if we should replace `https://www.example.gov` with `https://www.example.com` as that's a valid domain. Not a blocker and only a preference choice so happy for this to be merged without it.

gidjin (approved):
LGTM

PR #1031, opened 2023-05-30
fix: a11y: removes alt text from decorative images

shkeating:
# SC-2040

## Proposed changes

this branch removes alt text from decorative images as per Trusted Tester id [7.B - 1.1.1-decorative-image](https://section508coordinators.github.io/TrustedTester/images.html#for-decorative-images--check-111-decorative-image)

## Reviewer notes

decorative images should have empty alt text to prevent the non-sighted user from getting superflous info that may distract from the important page content.

## Setup

```sh
yarn dev
```

http://localhost:3000/

enter a search and check out the results page with ANDI to verify icon images of flag on moon and satellite don't have alt text


gidjin (approved):
LGTM

PR #1037, opened 2023-06-05
feat: a11y: assigns unique page titles to each page in the portal

shkeating:
# SC-2042

## Proposed changes

this branch uses [the Next API `<Head>` component](https://nextjs.org/docs/pages/api-reference/components/head) to assign unique values in the title element of each page. This ensures users that have multiple tabs of the portal open will see them uniquely labelled, and those using assistive tech can tell what page they are on.

This ensures the portal client passes Trusted Tester ID [2.4.2-page-title-purpose](https://section508coordinators.github.io/TrustedTester/titles.html#2.4.2-page-title-purpose)

## Reviewer notes
Is there page files living anywhere I didn't think of? I am opening a separate PR for the CMS in that repo.

you can also check page title in ANDI > structures > more details > page title

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

Navigate around and look at the page title output in the tab you see in your broswer, check that they are unique and that the names I use make sense.

---

### As a reviewer, I have

---

## Screenshots

image


gidjin (commented):
@abbyoung is going to take a look at this since we use `next/head` in our Header component to make sure it's not duplicating titles in the rendered HTML. Also we should update the generic error page too `src/pages/_error.tsx`

abbyoung:
We decided to pass `pageTitle` as a prop to all pages, except the error pages. The `404` and `500` pages can't have initial or server-side props, so the `<Head>` component works great. We kept the `<Head>` component in the generic `Error` page too to keep it consistent with that type of page.

For the rest of the pages, I took the existing title Shauna added in the `<Head>` component and passed it in from `getServerSideProps`. I also updated the corresponding tests.

Please let me know if I missed anything!

src/pages/404.tsx
```diff
@@ -21,6 +22,10 @@ export default function Custom404() {
     <Loader />
   ) : (
     <>
+      {/* 404 Page cannot have props, so we use the Head component to set the title */}
```
gidjin, line 25: thanks for the note!

src/pages/_app.tsx
```diff
@@ -33,12 +33,15 @@ type Props = AppProps & {
   hostname: {
     origin: string
   }
+  pageProps: Record<string, never>
```
gidjin, line 36: Question: Is this how you make the prop "required"?

src/pages/about-us/index.tsx
```diff
@@ -12,106 +12,109 @@ const AboutUs = () => {
   return !user ? (
     <Loader />
   ) : (
-    <div className={styles.aboutUsPage}>
-      <div className={styles.pageTitle}>
-        <h2>About the Space Force</h2>
-      </div>
+    <>
```
gidjin, line 15: nit-pick: Looks like this wrapper object is not really needed anymore. But It doesn't really hurt anything having it either.

src/pages/news-announcements.tsx
```diff
@@ -38,67 +39,75 @@ const NewsAnnouncements = ({
...
-    <div className={styles.listContainer}>
-      {announcements.length > 0 && (
-        <section>
-          <Announcement announcements={announcements} />
-        </section>
-      )}
+    <>
+      <Head>
+        <title>News & Annoucements - USSF Portal</title>
+      </Head>
```
gidjin, line 45: Suggestion: can this page use the `pageTitle` or is there some reason it needs to use the `Head` tag?

src/pages/update-browser.tsx
```diff
@@ -6,14 +6,17 @@ import Logo from 'components/Logo/Logo'
...
 export default function UpdateBrowser() {
   const MS_EDGE_DOWNLOAD = 'https://www.microsoft.com/edge'
   const GOOGLE_CHROME_DOWNLOAD = 'https://www.google.com/chrome'
   const FIREFOX_DOWNLOAD = 'https://www.mozilla.org/firefox'
 
   return (
     <>
+      <Head>
+        <title>Browser Warning - USSF Portal</title>
+      </Head>
```
gidjin, line 19: Suggestion: remove the `head` tag, looks like this page has both a head tag and a `pageTitle` prop.

gidjin (approved):
I found a couple pages that don't seem to use the new `pageTitle` attribute that I flagged. If you could fix them before merging that'd be great, but I don't see a need to re-review that.

src/pages/_app.tsx
abbyoung: It doesn't make it required, but makes TypeScript happy since we're now accessing the pageProps as an object in this file, instead of just passing it through to getLayout on line 176.

src/pages/about-us/index.tsx
```diff
@@ -139,7 +132,7 @@ AboutUs.getLayout = (page: React.ReactNode) =>
   )
 
 // The page title is parsed and displayed in _app.tsx
-export async function getServerSideProps() {
+export async function getStaticProps() {
```
gidjin, line 135: praise: Looks like this was a static page, good catch

gidjin (approved):

PR #1040, opened 2023-06-07
fix: Migration to remove duplicate users

jcbcapps:
# SC-2026

## Proposed changes
Adds migration to remove duplicate users. This is accomplished by:
- Merging the `mySpace` of each user into one
- If any of the duplicate users have a `dark` theme, we ensure that the final user has that value
- Keep the `displayName` of the original user
- Delete all duplicate users
- Finally, make the `userId` a unique field

---

## Reviewer notes
To test locally, you'll need to manually create some duplicate users at `http://localhost:8888/db/dev/users`
- Make sure that each user has the same `userId`

Try running `yarn migrate up` and refresh `http://localhost:8888/db/dev/users` (if `yarn migrate up` doesn't work, you may need to run `yarn migrate down` first):
- Any duplicate users should be consolidated into one, that contains a `mySpace` array featuring all collections from the deleted duplicates
- The remaining user should also only have one `FeaturedShortcut` and `GuardianIdeal`
- If any of the duplicate users has a theme set to `dark` we give the remaining user that value
- Try to add another user manually with the same `userId` and you should see an error thrown by Mongo. An example: `Something went wrong: BulkWriteError: E11000 duplicate key error collection: dev.users index: userId_1 dup key: { userId: "LINDSEY.WILSON.556070539@testusers.cce.af.mil" }`

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---


migrations/1686078533740-remove-duplicate-users.js
```diff
@@ -0,0 +1,55 @@
...
+      )
+    }
+  }
+
+  // Make the userId field unique
+  await db.collection('users').createIndex({ userId: 1 }, { unique: true })
+})
+
+module.exports.down = (next) => {
+  // Do nothing. We don't need to put the duplicate users back.
```
gidjin, line 53: Thought: Theoretically we could remove the unique index that is added in the down, but I'm not sure we want to do that since likely we should have had it from the start. Long way to say I'm good with the down of this being a no-op.

gidjin (approved):
Looks good to me. I ran this against a sample of production data and had no issues. I also spot checked a handful of records and they appear to have been merged properly.

abbyoung (approved):

PR #1045, opened 2023-06-15
ci: docker build status check

minhlai:
## Proposed changes

- Sometimes changes to dependencies will cause our build to break after the `e2e` stage. We currently do not have a test for the advanced stages of our Dockerfile that would warn us of these breaking changes before a PR is merged. This aims to solve that.

---

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---


.github/workflows/check-docker-build.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Build
+        uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671 # v4
+        with:
+          context: .
+          push: false
+          tags: portal-client:${{ github.sha }}
+          build-args: |
+            BUILD=${{ github.sha }}
+          cache-from: |
+            type=local,src=/tmp/portal-client,tag=e2e
```
gidjin, line 82: Question: Should tag here be `build` not `e2e`?

minhlai: I wasn't sure. I still want to use our saved build cache to optimize building the earlier stages (so our status checks don't take > 20 mins like they used to). But I wasn't sure if `tag=e2e` need to be the same tag as when the cache was saved (which was `tag=e2e`), like if I used a different tag to import than was used to export, it might not work. Since the build was working how I wanted, I left it rather than seeing if I could change it to `tag=build` https://github.com/USSF-ORBIT/ussf-portal-client/actions/runs/5280473472/jobs/9552648110

Willing to experiment though if you prefer we be explicit here!

gidjin (approved):

PR #1056, opened 2023-06-22
feat: Drag-and-drop entire collections

jcbcapps:
# SC-651

## Proposed changes
- Add drag-and-drop for collections
- Move most component logic from `MySpace` component into a new `MySpaceContext`

To do:
- ~Add test for `MySpaceContext`~

---

## Reviewer notes
Run the app as usual and test the following:
- Drag-and-drop using a mouse
- Drag-and-drop using a keyboard (tab to focus, space/enter to pick up/drop collections, arrow keys to move)
- Here is the e2e test to review: https://github.com/USSF-ORBIT/ussf-portal/pull/295

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---

## Screenshots

dragndrop


src/components/CustomCollection/EditableCollectionTitle.tsx
```diff
@@ -58,9 +59,17 @@ export const EditableCollectionTitle = ({
...
             placeholder={`Name this collection`}
             type="text"
+            onKeyDown={(e) => {
+              if (e.key === ' ') {
+                e.preventDefault()
+                const userInput = currentText + ' '
+                setCurrentText(userInput)
+              }
+            }}
+            onChange={(e) => setCurrentText(e.target.value)}
```
jcbcapps, line 72: The reason for adding this, is that when creating a new collection in My Space the user has to type a title for the collection, and if they hit 'space' it fires the `KeyboardSensor` for `dnd-kit`. This causes there to be no space added to the input field, and if the user hits the arrow keys the collection moves around. This is a hacky workaround, and I'm not sure why, but I don't like that I had to do this. It works as expected, but just feels strange.

src/types/index.ts
```diff
@@ -171,8 +171,14 @@ export type Widget = {
...
   title: string
   type: WidgetType
+  // Added this because the bookmarks field is required on the Collection interface below, but
+  // previously did not exist on the Widget type, which was causing a type error when trying to
+  // map over the user's MySpace to render the widgets.
+  bookmarks?: MongoBookmark[]
   // default indicates if the widget is automatically added for the user
   default?: boolean
+  // id is for dnd-kit
+  id?: string
```
jcbcapps, line 181: As the comment says, the difference in the presence of the `bookmarks` field was causing issues in a union type down the line. I believe we discussed this a few months back and a potential refactor of the MySpace-related types, but can't remember the decision. I added both `bookmarks` and `id` as optional fields, but let me know if that is potentially problematic and I'd be happy to take another look/create a new type.

src/components/MySpace/MySpace.tsx
```diff
@@ -138,159 +102,181 @@ const MySpace = ({ bookmarks }: { bookmarks: CMSBookmark[] }) => {
...
+          sensors={sensors}
+          collisionDetection={closestCorners}
+          onDragEnd={handleOnDragEnd}>
+          <Droppable dropId={'mySpaceDroppableArea'}>
+            <SortableContext
+              // Ignoring the following line because an id field has already been added to the Widget type to accomodate
+              // the unique identifier necessary for @dnd-kit.
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              items={mySpace}
```
jcbcapps, line 115: I opted to ignore this type for a few reasons. First, I already added an optional `id` field to the `Widget` type to accommodate the unique identifier. Second, the type error that shows here is complaining about the `MySpace` type and how it is not compatible with something like `(UniqueIdentifier | {id: UniqeIdentifier | undefined})[]` and I could not get that to line up with our current `MySpace` type. Third, even if I could get the types to line up, the `UniqueIdentifier` field would have to be required, and my initial thought it that we don't want that for `MySpace`. If you have any ideas/solutions, let me know!

src/components/ApplicationsTable/ApplicationsTable.tsx
```diff
@@ -17,9 +17,9 @@ type ApplicationsTableRowProps = {
 
 const ApplicationsTableRow = ({
   bookmark,
-  userCollectionOptions = [],
+  userCollectionOptions,
   handleAddToCollection,
-  canAddNewCollection = true,
```
jcbcapps, line 22: I was hunting for ways to bump test coverage and stumbled upon this. I removed these default params because they were never being evaluated due to the parent component having the same props with the same default param values.

src/components/CustomCollection/CustomCollection.tsx
```diff
@@ -177,11 +179,7 @@ const CustomCollection = ({
         'Save CMS link',
         `${title} / ${existingLink.label || existingLink.url}`
       )
-      handleAddBookmark(
-        existingLink.url || '',
-        existingLink.label,
-        existingLink.id
-      )
+      handleAddBookmark(existingLink.url, existingLink.label, existingLink.id)
```
jcbcapps, line 182: I changed this because `url` is a required value on the `CMSBookmark` type, so `existingLink.url` should never be false/empty

src/stores/myspaceContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+  mySpace: [],
+  disableDragAndDrop: false,
+  setDisableDragAndDrop: /* istanbul ignore next */ () => {
+    return
+  },
+  initializeMySpace: /* istanbul ignore next */ () => {
+    return
+  },
+  isCollection: /* istanbul ignore next */ () => {
+    return true || false
```
gidjin, line 42: Question: do these need to return `true || false` for the type system?

src/types/index.ts
gidjin: Is there a way we could use `_id` as the id field for dnd-kit? Or does it have to be `id`?

gidjin (approved):

jcbcapps:
> Is there a way we could use _id as the id field for dnd-kit? Or does it have to be id?

@gidjin I think you are right about this. The decision to do this may have just been a result of me working on this for a while and my brain burning out a little bit. I'm going to take another look at it.

src/stores/myspaceContext.tsx
jcbcapps: Yep!

shkeating (approved):
looks good to me, and passed all applicable trusted tester criteria. good work :) my only comment is that this is the kind of PR where adding a gif of the behavior to the screenshots section is helpful, but i went ahead and added it.

shkeating:
is chaging the contents of the myspace example in storybook on purpose?
image

jcbcapps:
> is changing the contents of the myspace example in storybook on purpose? image

@shkeating Nope. Looks like I forgot to mock the new `myspaceContext`. Good catch

PR #1060, opened 2023-07-05
dev: Hello world GraphQL API

jcbcapps:
# SC-2092

## Proposed changes
- Update docker-compose to add new personnel API service from our [new repo](https://github.com/USSF-ORBIT/ussf-personnel-api)

---

## Reviewer notes
After you've installed all dependencies, run `yarn services:up`, navigate to `http://localhost:4000/` and you should see the Apollo Server playground. Run the `hello` query.

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---


gidjin (commented):
This looks good, but the lint error is probably due to the fact that this is embedded. Continuing the discussion in slack to decide how we want to handle it.

docker-compose.services.yml
```diff
@@ -8,6 +8,16 @@ services:
     restart: always
     ports:
       - '6379:6379'
+
+  apollo-server:
+    container_name: personnel-api
+    build:
+      context: ../ussf-personnel-api/
```
gidjin, line 15: Suggestion: Having it this way means everyone, including our designers will need to have the repo checked out. I wonder if it would be better to pull from github by default and leave a comment on how to use the local copy for us devs who might be working on both repos at the same time. Thoughts? The below is an example similar to how the `cms` is setup in `docker-compose.cms.yml`, but a deviation from how we expect the cms to be running locally.

```suggestion
      # Uncomment the below line to work with local check out
      # context: ../ussf-personnel-api/
      context: https://github.com/USSF-ORBIT/ussf-personnel-api.git#main
```

gidjin (approved):

PR #1078, opened 2023-08-04
refactor: move portal user closer to session user in the code

gidjin:
## Proposed changes

Preparation for sending details of the user to matomo. 

## Reviewer notes

Originally I wanted to merge SessionUser and PortalUser but realized I couldn't. This is because one represents what is in the session sent from the SAML server and the other is what we store in the database. Both are not the same. I did move the data closer and reduced some redundancy in the code. Nothing should be different from a user perspective.

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

- [X] Created/modified automated [E2E tests](https://github.com/USSF-ORBIT/ussf-portal)

### As a reviewer, I have

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -3,7 +3,11 @@ import { Preview } from '@storybook/react'
 import { RouterContext } from 'next/dist/shared/lib/router-context'
 import { MockedProvider } from '@apollo/client/testing'
 import { ThemeProvider } from 'next-themes'
-import { UPDATE_GLOBALS, GLOBALS_UPDATED } from '@storybook/core-events'
+import {
+  UPDATE_GLOBALS,
+  GLOBALS_UPDATED,
+  FORCE_RE_RENDER,
+} from '@storybook/core-events'
 
 // happo support
 import { setThemeSwitcher } from 'happo-plugin-storybook/register'
@@ -151,6 +155,7 @@ const preview: Preview = {
   },
   decorators: [
     (Story, { globals: { theme } }) => {
+      console.log('theme provider decorator ', theme)
       return (
         <ThemeProvider
           forcedTheme={theme}
@@ -168,6 +173,7 @@ const preview: Preview = {
   ],
 }
 
+// this is only for happo
 setThemeSwitcher((theme: string, channel: any) => {
   return new Promise((resolve) => {
     // Listen for global to be updated and resolve.
@@ -176,6 +182,7 @@ setThemeSwitcher((theme: string, channel: any) => {
     channel.emit(UPDATE_GLOBALS, {
       globals: { theme },
     })
+    channel.emit(FORCE_RE_RENDER)
   })
 })
 
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -150,7 +150,19 @@ const preview: Preview = {
     },
   },
   decorators: [
-    (Story, { globals: { theme } }) => {
+    (Story, { globals: { theme }, parameters }) => {
+      // If only one happo theme is set then we want to only
+      // use that theme in the theme provider
+      if (parameters.happo.themes.length === 1) {
+        return (
+          <ThemeProvider
+            forcedTheme={parameters.happo.themes[0]}
+            enableSystem={false}
+            attribute={'data-color-theme'}>
+            <Story />
+          </ThemeProvider>
+        )
+      }
       return (
         <ThemeProvider
           forcedTheme={theme}
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -3,11 +3,7 @@ import { Preview } from '@storybook/react'
 import { RouterContext } from 'next/dist/shared/lib/router-context'
 import { MockedProvider } from '@apollo/client/testing'
 import { ThemeProvider } from 'next-themes'
-import {
-  UPDATE_GLOBALS,
-  GLOBALS_UPDATED,
-  FORCE_RE_RENDER,
-} from '@storybook/core-events'
+import { UPDATE_GLOBALS, GLOBALS_UPDATED } from '@storybook/core-events'
 
 // happo support
 import { setThemeSwitcher } from 'happo-plugin-storybook/register'
@@ -155,7 +151,6 @@ const preview: Preview = {
   },
   decorators: [
     (Story, { globals: { theme } }) => {
-      console.log('theme provider decorator ', theme)
       return (
         <ThemeProvider
           forcedTheme={theme}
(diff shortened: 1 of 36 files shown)


src/stores/analyticsContext.test.tsx
```diff
@@ -44,13 +43,18 @@ window.fetch = jest.fn(() =>
...
+// ignore console.debug and don't output to test output
+jest.spyOn(console, 'debug').mockImplementation(jest.fn())
+const mockConsoleWarn = jest.fn()
+jest.spyOn(console, 'warn').mockImplementation(mockConsoleWarn)
+
 describe('useAnalytics', () => {
-  it('throws an error if AnalyticsContext is undefined', () => {
+  test('throws an error if AnalyticsContext is undefined', () => {
     jest.spyOn(React, 'useContext').mockReturnValueOnce(undefined)
     expect(() => useAnalytics()).toThrowError()
```
gidjin, line 54: I suspect this test is not testing what it is expecting. Also running this suite results in console log below that leads me to believe the tests here need to be refactored. But this PR is already large enough so I left it for now. Would love thoughts on this.

```sh
  console.error
    Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.

      158 |
      159 | export const useAnalytics = () => {
    > 160 |   const context = useContext(AnalyticsContext)
          |                             ^
      161 |
      162 |   if (context === undefined) {
      163 |     throw new Error(

      at printWarning (node_modules/react/cjs/react.development.js:209:30)
      at error (node_modules/react/cjs/react.development.js:183:7)
      at resolveDispatcher (node_modules/react/cjs/react.development.js:1592:7)
      at useContext (node_modules/react/cjs/react.development.js:1602:20)
      at useAnalytics (src/stores/analyticsContext.tsx:160:29)
      at src/stores/analyticsContext.test.tsx:54:30
      at Object.<anonymous> (node_modules/expect/build/toThrowMatchers.js:74:11)
      at Object.throwingMatcher [as toThrowError] (node_modules/expect/build/index.js:312:21)
      at Object.toThrowError (src/stores/analyticsContext.test.tsx:54:34)

 PASS  src/stores/analyticsContext.test.tsx
  useAnalytics
    ✓ throws an error if AnalyticsContext is undefined (28 ms)
    ✓ returns the created context (6 ms)
  Analytics context
    ✓ warns in the console if missing the analytics URL (55 ms)
    ✓ warns in the console if missing the analytics site ID (54 ms)
    ✓ initializes the analytics queue on render (5 ms)
    with config defined
      ✓ initializes the analytics config on mount (4 ms)
      ✓ provides the push handler (41 ms)
      ✓ provides the trackEvent handler (15 ms)
      ✓ can use trackEvent to track an event name (15 ms)
      ✓ can use trackEvent to track an event value (14 ms)

```

src/components/ThemeToggle/ThemeToggle.stories.tsx
```diff
@@ -29,10 +31,25 @@ export default {
...
+          }),
+        },
+      ]
       return (
         <AuthContext.Provider value={mockContext}>
-          <Story />
+          <MockedProvider mocks={mocks}>
+            <Story />
+          </MockedProvider>
         </AuthContext.Provider>
```
gidjin, line 53: These changes were necessary for storybook to show the correct text for the theme toggle button when the theme was changed. This has the side effect of refreshing the entire view, including the storybook manager. It's awkward, but I'm not sure there is another way to properly support it. One option might be to switch to having a dark them and light theme sub story but that would be a big change on how we use storybook currently. Would love to chat about this at a future sync, or here and see what others have to say / think on this.

src/stores/analyticsContext.test.tsx
abbyoung: my understanding is the console error is what we expect from that test -- if the hook is used improperly, it'll throw that error. we have a similar function/test in the other contexts as well. maybe i'm missing something?

jcbcapps: Yeah I agree with @abbyoung that the console error looks to be expected behavior, but the error message in `useAnalytics` is 'useAnalytics must be used within an AnalyticsContextProvider' and I'm not seeing that anywhere in the output. So maybe that is an issue? Maybe we should be seeing that specific error message as opposed to the bulleted list in the above read out.

jcbcapps:
@gidjin @abbyoung I'm seeing this error. Not sure what it is, but it's popping up before I can get to the login page
Screenshot 2023-08-21 at 10 04 59 AM

.storybook/preview.tsx
```diff
@@ -177,6 +178,8 @@ setThemeSwitcher((theme: string, channel: any) => {
     channel.emit(UPDATE_GLOBALS, {
       globals: { theme },
     })
+    // Force re-render of all stories
```
abbyoung, line 181: Testing to see if this fixes the happo render issues

shkeating:
I think we got tagged in here bc of a happo malfunction that is not a front end change, so unless told to do otherwise, I will be approving this o

shkeating (approved):
Approving to prevent me from being a blocker

abbyoung:
> @gidjin @abbyoung I'm seeing this error. Not sure what it is, but it's popping up before I can get to the login page Screenshot 2023-08-21 at 10 04 59 AM

@jcbcapps this happened for me and unfortunately i can't remember the root cause..i believe it was my local environment. let's pair tomorrow?

src/stores/analyticsContext.test.tsx
gidjin: iirc I tried to make the test be more specific by looking at what error was returned and had trouble getting it to work. I was hoping to avoid printing expected errors to the logs in our tests since it's frequently a source of confusion for me. But I think this was one I couldn't fix. I'm leaving it as is for now :)

.storybook/preview.tsx
```diff
@@ -151,7 +150,19 @@ const preview: Preview = {
     },
   },
   decorators: [
-    (Story, { globals: { theme } }) => {
+    (Story, { globals: { theme }, parameters }) => {
+      // If only one happo theme is set then we want to only
+      // use that theme in the theme provider
+      if (parameters.happo.themes.length === 1) {
```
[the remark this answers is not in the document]
gidjin: Thanks for flagging, it crossed my mind but I forgot to come back and verify if I needed to check for nulls.

abbyoung (approved):

jcbcapps (approved):
LGTM

PR #1079, opened 2023-08-07
feat: Create weather widget

jcbcapps:
# SC-1444

## Proposed changes
- Add Weather Widget to My Space

TO DO:
- ~Unit tests~
- ~E2E tests~
- ~Add documentation for adding a new widget~
- ~Update Storybook component to use mock data~
- ~Add LD flag~

---

## Reviewer notes
All you need to do in order to test:
- Click the 'Add widget' component in My Space and select 'Add weather widget'
- You should see a new widget appear, prompting you to enter a zip code
- After saving the zip code, you should see the weather appear
- You can click the settings icon in the top right to edit/remove
- You can also click the zip code itself to edit
- You should also be able to drag-and-drop weather widgets

Currently, since the rate limit for the National Weather Service API is not public knowledge, I have limited the number of allowed weather widgets to 3. That is semi-arbitrary and I welcome feedback on that decision.

Check out the e2e tests: https://github.com/USSF-ORBIT/ussf-portal/pull/308

## Setup
IMPORTANT:
In order for this to work, you need to have the `sc-2156-backend-weather-widget` branch checked out in the CMS repo (at least until it's merged into `main`)

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---

## Screenshots
one
two
three
four
five


src/components/WeatherWidget/WeatherWidget.tsx
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  const handleKeyPress = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    const keyCode = e.key
+
+    // Allow only digits (0-9)
+    if (!/^\d$/.test(keyCode)) {
+      e.preventDefault()
+    }
+  }
```
jcbcapps, line 142: I had to add this because, even after adding a pattern to the input and restricting it to 5 digits, you could still type special characters for some reason. So this function intercepts each key press, and if it isn't a single digit, it will not allow the value to be placed in the input field.

shkeating:
Hmmm I'm personally not vibing with the black icons on light mode, going to discuss on slack with Mal on what we would prefer

jcbcapps:
> Hmmm I'm personally not vibing with the black icons on light mode, going to discuss on slack with Mal on what we would prefer

Agreed. That's just what is happening by default but I would definitely love some input on an alternative! I meant to mention this at team sync but forgot.

shkeating:
> > Hmmm I'm personally not vibing with the black icons on light mode, going to discuss on slack with Mal on what we would prefer
> 
> @shkeating Agreed. That's just what is happening by default but I would definitely love some input on an alternative! I meant to mention this at team sync but forgot.

I'm trying to pull the branch and run it but it won't let me submit a zip code to see what it looks like

jcbcapps:
> > > Hmmm I'm personally not vibing with the black icons on light mode, going to discuss on slack with Mal on what we would prefer
> > 
> > 
> > @shkeating Agreed. That's just what is happening by default but I would definitely love some input on an alternative! I meant to mention this at team sync but forgot.
> 
> I'm trying to pull the branch and run it but it won't let me submit a zip code to see what it looks like

@shkeating Do you have the `sc-2156-backend-weather-widget` branch checked out in the CMS repo?

shkeating (changes requested):

shkeating:
I just set the colors up for the icons in dark & light mode

shkeating:
Things a looking good but I need to do accessibility review still

shkeating:
## Accessibility Issue - Heading Order

The use of h1s and h3s are resulting in breaking the header order. Since this information falls below the h3 of the Weather Widget we should not use heading level 3 or up inside of it. We make want to use h4/h5 semantically but style them to _look_ like our h1 and h3.

image

shkeating:
## Usability Issue: Unable to press keys that are not numbers

This includes pasting a ZIP and backspace keys to edit not working.

shkeating:
## Usability Issue - Identifying location of weather widget

In our designs, we had planned for the name of the location to be output onto the widget. I somewhat remember y'all saying that might not be possible with the API. Would it be possible to follow a pattern similar to collections where they can add a nickname or something. I foresee it being hard to determine where the ZIP goes to if the person doesn't have it memorized, or they even have more than one weather widget in their My Space mixing them up quite easily.

If it can only be added via customization/user input , I would be inclined to say that is out of scope for the initial release of this widget, and we should add a backlog item to figure out a way to identify weather widgets.

image

design:
image

jcbcapps:
> ## Usability Issue - Identifying location of weather widget
> In our designs, we had planned for the name of the location to be output onto the widget. I somewhat remember y'all saying that might not be possible with the API. Would it be possible to follow a pattern similar to collections where they can add a nickname or something. I foresee it being hard to determine where the ZIP goes to if the person doesn't have it memorized, or they even have more than one weather widget in their My Space mixing them up quite easily.
> 
> If it can only be added via customization/user input , I would be inclined to say that is out of scope for the initial release of this widget, and we should add a backlog item to figure out a way to identify weather widgets.
> 
> image
> design: image

@shkeating Getting the name of the location is not currently possible with the API. As you suggested, it would be possible to give users the ability to add a custom name for a location, but I would have to agree with you that that is outside of the scope of this story. That would entail modifying both front end and back end code. If that is what we want to do, we might want to consider doing it before we release the widget to users, because I agree that it will be difficult for some to remember the location by zip.

src/components/WeatherWidget/WeatherWidgetIcon.test.tsx
```diff
@@ -0,0 +1,485 @@
...
+
+    await act(async () => {
+      renderWithMySpaceAndModalContext(
+        <WeatherWidgetIcon hourlyPeriod={mockHourlyPeriod} iconSize="2xl" />,
+        {}
+      )
+    })
+
+    await waitFor(() => {
+      expect(screen.getByLabelText('Icon for cloud')).toBeInTheDocument()
```
gidjin, line 314: Question: are there different icons for cloudy or rain that are different for night or day? I don't think so but wanted to confirm.

gidjin (approved):

gidjin:
I'm not sure about the limit of 3 weather widgets. That seems fine, I'd be curious if folks request more.

src/components/WeatherWidget/WeatherWidgetIcon.test.tsx
jcbcapps: In font awesome? Not that I'm aware of. But importing into the app? No, we're using the same ones.

jcbcapps:
> I'm not sure about the limit of 3 weather widgets. That seems fine, I'd be curious if folks request more.

Yeah I chose that number somewhat arbitrarily because we don't know the rate limit of the NWS api. I figured it would be better to start out offering less, than allowing for too many and users start to see failed api requests. Curious to see how it performs once we get it in front of users.

abbyoung:
If I try to add an invalid zip code, or if there's an error in fetching, currently nothing happens in the UI for me. Do we want to surface an error to the user so they know something went wrong? 

My only concern with limiting to 3 weather widgets is how we communicate that to users -- will it be confusing if they can no longer add any more weather widgets? Maybe more of a question for @malworks and @shkeating.

Screenshot 2023-08-23 at 9 00 17 AM

shkeating:
my only notes now are I think the text in the hourly forecast should be aligned center
image

src/components/WeatherWidget/WeatherWidget.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+export default {
+  title: 'Components/WeatherWidget',
+  component: WeatherWidget,
+} as Meta
+
+export const AddWeatherWidget = () => <WeatherWidget />
+
+export const WeatherWidgetWithData = () => (
+  <WeatherWidget widget={mockWeatherWidget} />
+)
```
gidjin, line 47: Suggestion: We probably should mock out the api call to weather.gov in storybook like we do with the RSS widget. Otherwise we'll have a happo diff every time the weather changes. 

Question: should we add mocks to show some of the other weather possibilities?

gidjin (changes requested):
Requesting change to mock out the weather widget with data in storybook, if we can't let me know and I'll approve.

shkeating (approved):
with the hourly forecast centered its lookin good to me!

shkeating:
> Requesting change to mock out the weather widget with data in storybook, if we can't let me know and I'll approve.

+1 to this tho. pretty sure it makes it hard to run storybook by itself if you don't use mock data

src/components/WeatherWidget/WeatherWidget.stories.tsx
gidjin: I added code to do this as discussed.

gidjin (approved):

shkeating (approved):

PR #1092, opened 2023-08-31
dev: Set EDIPI as userId in Matomo

jcbcapps:
# SC-2270

## Proposed changes
Associate the user's edipi that we get from SAML with a profile in Matomo

---

## Reviewer notes
Once you have everything running locally:
- Navigate to Matomo at `http://localhost:8081/` and log in
- In another tab, navigate to the portal and log in, take note of the user's edipi which should be the numbers in their userId. This is what you should see in Matomo
Screenshot 2023-08-31 at 11 39 10 AM
- After logging in, look at the dashboard in Matomo and find the User IDs section under the Visitors tab
- You should see the edipi of the user that you are logged in with. Hover over the id and you should see a little green box like in the screenshot below. Click that and you should see their profile
Screenshot 2023-08-31 at 11 24 59 AM
- Here is what the user's visitor profile should look like:
Screenshot 2023-08-31 at 11 23 10 AM

Something you will probably see while in Matomo is the following warning. I looked into this and it doesn't seem to have any negative effect on this PR. There is already a fix for this coming in Matomo v5 ([source](https://forum.matomo.org/t/deprecation-error-in-plugins-userid-reports-getusers-php/51551/6)):
Screenshot 2023-08-31 at 11 15 17 AM

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---


src/stores/authContext.tsx
```diff
@@ -34,6 +35,7 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
   const [user, setUser] = useState<SessionUser | null>(null)
   const [portalUser, setPortalUser] = useState<PortalUser | null>(null)
   const router = useRouter()
+  const { setUserIdFn, unsetUserIdFn } = useAnalytics()
```
gidjin, line 38: Question: what do you think about mocking these out so we can add a test to ensure they are called in `authContext.test.tsx`? Is that something you tried? Maybe the value of that is low?

gidjin (commented):
Looks good, though I only saw the tracking in the console, stuff didn't show up in matomo. I'll check again later if it needs more time. That said I did have issues with matomo locally so maybe @jbecker01 or @abbyoung can confirm it works for them.

jbecker01:
Unsure if this matters but in the description, it lists Matomo being on port 8081 (which I could not get to work) but in the dev readme it says 8443 (which I could get to work)

src/stores/authContext.tsx
jcbcapps: @gidjin I have some tests for `analyticsContext` that are calling each of these functions and verifying that by looking at console output. I'm not sure how I would go about mocking that out for the `authContext` to verify the same. As for the level of value that would bring, I'm not sure? The code coverage report is showing that the functions are reached in `authContext`, and the tests in `analyticsContext` are checking the same events that would be sent to Matomo when the login/logout functions run. The combination of those two seem fine to me, but curious to know if others agree?

jcbcapps:
> Unsure if this matters but in the description, it lists Matomo being on port 8081 (which I could not get to work) but in the dev readme it says 8443 (which I could get to work)

@jbecker01 were you able to see the user id in Matomo on port 8443? There was a recent update and it should be accessible on 8081 on your local machine. Not sure if both ports are supposed to work locally.

jbecker01:
> > Unsure if this matters but in the description, it lists Matomo being on port 8081 (which I could not get to work) but in the dev readme it says 8443 (which I could get to work)

8081 still isn't working for me. On 8443, when I click on "portal" under all websites, I get redirected back to the tracking code page, not the analytics dashboard as expected.

jcbcapps:
@jbecker01 that's strange. Maybe we could pair on this today if you have the time?

jbecker01:
After Jacob's help, I was finally able to test this and it looks good!

src/stores/authContext.tsx
gidjin: @jcbcapps I added some expectations to the authContext. If they look good to you I'm good with merging this in.

gidjin (approved):

PR #1097, opened 2023-09-11
fix: use date from api for weather widget

gidjin:
## Proposed changes

The weather widget was using the current date for displaying the month and day in the weather widget. This PR changes it to use the date in the forecast response. Also this PR changes the unit test to mock the call to the api.

Also this PR uses the special attribute `data-happo-hide` on the star field in the loader as described in the [happo docs on spurious diffs](https://docs.happo.io/docs/spurious-diffs) to hide the star field from happo. The star field is randomly generated each build so it will always be different in each build.

## Reviewer notes

The fixtures used have a hard coded date of Aug 16th, 2023 which was a Wednesday. This is reflected in the unit test conditions I added and in the storybook renderings, see screenshot below.

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

### Start storybook

```sh
yarn storybook
```

---

### As a reviewer, I have

---

## Screenshots

SCR-20230911-ezm

### DIFF
--- a/src/components/Loader/Loader.tsx
+++ b/src/components/Loader/Loader.tsx
@@ -13,7 +13,7 @@ const Loader = () => (
       <div className={styles.back}>
         <div className={styles.earth}></div>
         <div className={styles.cosmos}>
-          <div className={styles.star}></div>
+          <div className={styles.star} data-happo-hide></div>
           <div className={styles.spinner}>
             <div className={styles.face}>
               <div className={styles.circle}></div>
--- a/src/components/WeatherWidget/WeatherWidget.test.tsx
+++ b/src/components/WeatherWidget/WeatherWidget.test.tsx
@@ -1,11 +1,13 @@
 /**
  * @jest-environment jsdom
  */
+import axios from 'axios'
 import { act, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import React from 'react'
 import { ObjectId } from 'mongodb'
 import { renderWithMySpaceAndModalContext } from '../../testHelpers'
+import { mockHourlyForecast } from '__fixtures__/data/hourlyForecast'
 import WeatherWidget from './WeatherWidget'
 import { WeatherWidget as WeatherWidgetType } from 'types'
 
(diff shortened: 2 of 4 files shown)


src/components/WeatherWidget/WeatherWidget.test.tsx
```diff
@@ -41,6 +51,10 @@ const mockWeatherWidgetWithIncorrectZipcode: WeatherWidgetType = {
 }
 
 describe('WeatherWidget', () => {
+  mockedAxios.get.mockImplementation(() => {
+    return Promise.resolve({ data: mockHourlyForecast })
+  })
+
```
jcbcapps, line 57: I think this might be why that one unit test is failing. The test takes `mockWeatherWidgetWithIncorrectZipcode` which intentionally contains an invalid url, but I think this mocked api call is giving the test data. Maybe you need to just move the test into another `describe`?

jcbcapps (approved):
LGTM! Left a note on the failing test, but everything else looks good.

src/components/WeatherWidget/WeatherWidget.test.tsx
gidjin: Thanks for the tip, this made it easier to find!

src/components/Loader/Loader.tsx
```diff
@@ -13,7 +13,7 @@ const Loader = () => (
       <div className={styles.back}>
         <div className={styles.earth}></div>
         <div className={styles.cosmos}>
-          <div className={styles.star}></div>
+          <div className={styles.star} data-happo-hide></div>
```
abbyoung, line 16: Q: Should this impact the happo diffs on this PR, or in future PRs?

abbyoung (approved):
LGTM! Left one non-blocking question to better understand happo behavior.

src/components/Loader/Loader.tsx
shkeating: I think we should ignore the loader in Happo. It has very tightly scoped CSS and it is unlikely it would be affected by other styling inadvertently. So hopefully, this gets it to stop showing up. It is getting caught every time because the stars in the background generate randomly so its capturing a new image every run 🤦‍♀️

shkeating (approved):
voting yes to skipping the loader in happo. yay for mocked data date!!

shkeating:
oh hiding the stars from happo works too. I think that approach would work.

src/components/Loader/Loader.tsx
gidjin: I expect happo to have a diff on the loader this time, one with a star field from before and mine without. But in future PRs the star field won't display for happo at all so shouldn't cause a diff.

PR #1120, opened 2023-10-13
ci(Mergify): configuration update

minhlai:
This change has been made by @minhlai from Mergify config editor.


gidjin (approved):
LGTM, we need to sign the commit though which Mergify Config Editor seems to not do for us.

PR #1144, opened 2023-11-03
feat: redirectTo support

gidjin:
# SC-2668

## Proposed changes

Continuation of work started in https://github.com/USSF-ORBIT/ussf-portal-client/pull/579. This PR introduces server side support for the `redirectTo` parameter. This parameter is captured, if it is present on the login page, and passed through the SAML authentication using the SAML property `RelayState`. This is then passed back to us when the SAML server posts to our callback end point. 

If the parameter is present during the call back the user is redirected to the destination.

To prevent miss use of this feature, the `redirectTo` field only supports relative urls. So for example `/new-announcements`. The values that start with `/cms` are considered special. Before we redirect the `/cms` portion will be replaced with the value of the environment variable `KEYSTONE_PUBLIC_URL`. This allows us to redirect users to the CMS without allowing absolute urls.

If a user goes to a url other than / or /login the path is captured automatically and passed to the login flow under `redirectTo` query parameter.

---

## Reviewer notes

I found we were redirecting to the login page twice when a user was not logged in. Once when the `authContext` found that the user was not logged in, and once when the graphql client found the same issue. This was preventing the `redirectTo` from being properly preserved. I changed the graphql client to just log the error and rely on the authContext for the redirect. Let me know if this is an issue.

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

1. Test that logging in to the CMS first will redirect you there
  - http://localhost:3001
  - http://localhost:3001/users
2. Test that logging in to a specific page redirects you back
  - http://localhost:3000/news-announcments
3. Test that logging in as normal works
  - http://localhost:3000
  - http://localhost:3000/login

### Start storybook

```sh
yarn storybook
```

---

- [X] Met the acceptance criteria
- ~[ ] Created new stories in Storybook if applicable~
- [X] Created/modified automated [E2E tests](https://github.com/USSF-ORBIT/ussf-portal)
- [X] Followed guidelines for zero-downtime deploys, if applicable
- ~[ ] Use [ANDI](https://www.ssa.gov/accessibility/andi/help/install.html) to check for basic a11y issues~

### As a reviewer, I have


src/apolloClient.tsx
```diff
@@ -16,7 +16,8 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
       )
       switch (extensions.code) {
         case 'UNAUTHENTICATED':
-          window.location.href = '/login'
+          console.error('user not logged in')
+          //window.location.href = '/login'
```
gidjin, line 20: This line was causing us to redirect to `/login` twice, once here and once from the `useEffect` in the `authContext`. That was causing a race condition when adding the `redirectTo` flag. Changing this resolved that but I'm not sure if this creates some other issue.

abbyoung (approved):
LGTM! Works great!

PR #1169, opened 2023-11-22
feat: Adds new graphql server for thirdparty api

abbyoung:
# SC-2681

### Background
This PR is the result of some exploratory coding to determine the best implementation of a v1 third-party API. I've operated under the following assumptions:

* The first query to this API will be for articles tagged 'C-Note', and will be publicly accessible.
* The API should support both authorized and unauthorized (public) queries, although v1 only includes a public query.
* The API should be low-lift and able to get up and running quickly, with the flexibility to expand in the future.

### Design
This approach uses the built-in api routes supplied by Next.js in tandem with Apollo Server. It follows the same pattern as our internal GraphQL API, except it's placed in a different directory to avoid collisions (`api/thirdparty/graphql.tsx`). While I haven't implemented tests yet, this seems to work fine. The existing API remains functional, and the new API remains accessible even if a user is logged out (unauth/public).

There are two key components to the API:

`pages/api/thirdparty` - This directory contains all the files needed to start and run the GraphQL server: typeDefs, resolvers, server, queries, data sources.

~~`pages/api/dataSources/keystone.ts` - This file already existed as a dataSource used by the internal GraphQL API. Since this first query is for the Keystone API, it made sense to add the function/query to the dataSource and continue to follow this existing pattern. `getArticles` takes the optional parameter of `tag`, which will allow for querying C-Notes.~~ On second/third thought, I've split out this code into a separate `ThirdPartyKeystone` dataSource under the `pages/api/thirdparty` directory -- I can see how it will quickly become confusing to have the code too intertwined. This will also make it easier for splitting into a separate service in the future (cc @gidjin @jcbcapps  let me know what you think)

### Future Work
* Apollo Server makes it possible to add authorization to specific queries/resolvers, when we're ready for that functionality. Docs here: https://www.apollographql.com/docs/apollo-server/security/authentication/#authorization-methods
* Rate limits: If/when the API begins to receive more traffic than our app can reasonably support (in addition to supporting the portal), we'll need to consider pulling out the server code into its own service. This should be relatively straightforward.

### Start the system
```sh
yarn services:up
yarn dev
cd ../ussf-portal-cms
yarn dev
```

You _do not_ have to log in.

Go to http://localhost:3000/api/thirdparty/graphql

Run the following query:
~~UPDATE 11/28 -- New type for $tag and enum value to pass in as variable (cc @gidjin @jcbcapps since you've already reviewed with the older query)~~
UPDATE 11/30 -- No longer require a tag, the query will only return cNotes

```
  query GetCNotes {
    cNotes
      {
        id
        title
        publishedDate
       	body {
          document
         }
    }
  }
```

Note: Make sure you've created an article with the category InternalNews and tagged it 'C-Note' (per the docs guidance on creating CSO notices).

### As a reviewer, I have

---


src/operations/cms/queries/getInternalNewsArticles.ts
```diff
@@ -1,12 +1,13 @@
 import { gql } from 'graphql-tag'
 
 export const GET_INTERNAL_NEWS_ARTICLES = gql`
-  query GetInternalNewsArticles($publishedDate: DateTime) {
+  query GetInternalNewsArticles($publishedDate: DateTime, $tag: String) {
```
abbyoung, line 4: By adding $tag as an optional param, it allowed us to reuse the existing query and keep things DRY

src/operations/cms/queries/getInternalNewsArticles.ts
```diff
@@ -20,6 +21,12 @@ export const GET_INTERNAL_NEWS_ARTICLES = gql`
         name
         type
       }
+      body {
```
abbyoung, line 24: We need to return the body of the article here, since this particular query is interested in getting the entire article and not just the preview

src/pages/api/thirdparty/graphql.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { DateTime } from 'luxon'
+import { startServerAndCreateNextHandler } from '@as-integrations/next'
+import { GraphQLJSON } from 'graphql-type-json'
+import { GraphQLError } from 'graphql'
+import KeystoneAPI from '../dataSources/keystone'
+
+/* Type Definitions */
+
+/* Data we will return to any third-party querying the API. */
+const typeDefs = `
```
abbyoung, line 11: It's a bit repetitive to define these typeDefs when these are already defined in the Keystone CMS GraphQL API. We could potentially import the entire schema from Keystone and use that for typeDefs, but then we'd have to manually update it every time a change was made in Keystone, and also would be limited to returning only the data that Keystone exposes.

src/pages/api/thirdparty/graphql.tsx
```diff
@@ -0,0 +1,87 @@
...
+  scalar JSON
+
+  type Article {
```
abbyoung, line 14: TODO: The data we return may need to be updated based on requirements from third-party

src/pages/api/thirdparty/graphql.tsx
```diff
@@ -0,0 +1,87 @@
...
+    id: ID!
+    title: String!
+    publishedDate: String!
+    tag: String!
+    body: Article_body_Document
+  }
+
+  type Article_body_Document {
```
abbyoung, line 22: This mirrors the schema from Keystone

src/pages/api/thirdparty/schema.ts
```diff
@@ -0,0 +1,22 @@
+import { gql } from 'graphql-tag'
+
+/* Data we will return to any third-party querying the API. */
```
abbyoung, line 3: It's a bit repetitive to define these typeDefs when these are already defined in the Keystone CMS GraphQL API. We could potentially import the entire schema from Keystone and use that for typeDefs, but then we'd have to manually update it every time a change was made in Keystone, and also would be limited to returning only the data that Keystone exposes.

src/pages/api/thirdparty/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  type Article {
+    id: ID!
+    title: String!
+    publishedDate: String!
+    tag: String!
+    body: Article_body_Document
+  }
+
+  type Article_body_Document {
```
abbyoung, line 15: This mirrors the schema from Keystone

src/operations/cms/queries/getInternalNewsArticles.ts
gidjin: Thought: I wonder if we'll want to reuse this in the future or if there will be more drift in the queries. This works since it's low lift and we don't know how things will look once folks start using the api.

src/pages/api/thirdparty/schema.ts
gidjin: Thought: I think it's fine to have a little repetition. While the typeDefs are similar now they may not be in the future and the audience of each is different.

src/pages/api/thirdparty/resolvers.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export type ThirdPartyContext = {
+  dataSources: {
+    keystoneAPI: KeystoneAPI
+  }
+}
+
+/* Resolvers */
+export const resolvers = {
```
gidjin, line 18: Question: If I'm reading this right tag is a parameter passed in, where would we be confirming that third parties can only request `c-note` since as of now the API is public? Or is that something that is coming soon?

abbyoung: GREAT catch. I need to address that.

src/operations/cms/queries/getInternalNewsArticles.ts
abbyoung: This is good feedback. Paired with some other changes, I decided it's worth making this switch now and creating a new query under the thirdparty directory. Keep all the third-party code together, instead of getting too intertwined with existing code. Lmk if you have any concerns!

src/operations/cms/queries/getInternalNewsArticles.ts
abbyoung: update: removing this, moving query into separate file

src/pages/api/thirdparty/resolvers.ts
abbyoung: Update: I've changed tag from String to an Enum, and added CNOTE as one of the enum values.

src/pages/api/thirdparty/schema.ts
```diff
@@ -0,0 +1,26 @@
+import { gql } from 'graphql-tag'
+
+/* Data we will return to any third-party querying the API. */
+export const typeDefs = gql`
+  scalar JSON
+
+  enum ArticleTag {
```
abbyoung, line 7: @gidjin @jcbcapps What do y'all think about creating a group of enums for approved tags to be passed to the getPublicArticles query?

src/pages/api/thirdparty/dataSources/thirdPartyKeystone.ts
```diff
@@ -0,0 +1,36 @@
+import { RESTDataSource } from '@apollo/datasource-rest'
+import type { KeyValueCache } from '@apollo/utils.keyvaluecache'
+import { DateTime } from 'luxon'
+import { print } from 'graphql'
+import { GET_PUBLIC_ARTICLES } from '../operations/getPublicArticles'
+
+export const ARTICLE_PUBLIC_TAGS = {
+  CNOTE: 'C-Note',
```
gidjin, line 8: Thought: Since the enum value is `CNOTE` that's what I made the tag name before I realized that the tag needs to be `C-Note`. I like the use of an enum to control what values can be requested, and since this is just the first version I don't know that we need to change it. But my wondering is can we make the enum name and value match? I'll bring this up at eng sync too since I think a sync discussion might be good for this.

src/pages/api/thirdparty/dataSources/thirdPartyKeystone.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+class ThirdPartyKeystoneAPI extends RESTDataSource {
+  override baseURL = process.env.KEYSTONE_URL
+
+  constructor(options: { cache: KeyValueCache }) {
+    super(options)
+  }
+
+  async getPublicArticles(publishedDate: DateTime, tag?: string) {
```
gidjin, line 18: Question: I'm missing how this works when you don't pass in the tag. Does apollo client return all data across all the enum values? We only have one so it isn't immediately clear to me.

src/pages/api/thirdparty/resolvers.test.ts
```diff
@@ -0,0 +1,106 @@
...
+            type: string
+            children: {
+              text: string
+            }[]
+          }[]
+        }
+      }[]
+    }
+
+    test('returns all articles tagged with C-Note', async () => {
```
gidjin, line 49: Suggestion(non-blocking): For a test that checks the negative case, ie they pass something that isn't `CNOTE` and maybe one without. This isn't blocking given we are in the early stage of figuring out what our users will need.

src/pages/api/thirdparty/schema.ts
gidjin: I'm good with this approach. My only question, which is in another comment, is around the fact that the API will take `CNOTE` and the tag is actually `C-Note`.

gidjin (commented):
Praise: Thanks for making the changes, I like how much simpler everything seems.

src/pages/api/thirdparty/schema.ts
```diff
@@ -4,23 +4,18 @@ import { gql } from 'graphql-tag'
...
-  enum ArticleTag {
-    CNOTE
-  }
-
-  type Article {
+  type CNote {
     id: ID!
     title: String!
     publishedDate: String!
-    tag: ArticleTag!
```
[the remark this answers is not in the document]
abbyoung: I feel like we'll need to add / edit fields once the API is tested by others, so maybe leave until future iteration

src/pages/api/thirdparty/operations/getPublicArticles.ts
```diff
@@ -0,0 +1,32 @@
...
+      where: {
+        status: { equals: Published }
+        publishedDate: { lte: $publishedDate }
+        category: { equals: InternalNews }
+        tags: { some: { name: { equals: "C-Note" } } }
+      }
+      orderBy: [{ publishedDate: desc }]
+    ) {
+      id
+      slug
```
jcbcapps, line 15: I noticed that `slug` isn't being returned in the api response, and I'm wondering if we need it?

jcbcapps (approved):
LGTM! Great work. One call out I have is that in the API response, there is no distinction between the preview text and the article body. I don't think this will be an issue, but just wanted to flag it. 
Screenshot 2023-12-01 at 11 45 17 AM

PR #1193, opened 2024-01-10
refactor: Initial commit - duplicating Dockerfile and pointing gh actions to it

CoryAtTruss:
# SC-3028
Duplicated the Dockerfile to allow for future branching with the mirroring to gitlab - the default will point to IronBank images, and the new Dockerfile-gh will remain the same. We have more flexibility with GH Actions and where it pulls from, so the default will be what is used in Gitlab.

This PR also attempts to start pointing our GH Actions at the new Dockerfile (which for now, is exactly the same).


docker-compose.cms.yml
```diff
@@ -5,7 +5,7 @@ services:
       # Uncomment the below line to work with local check out
       # context: ../ussf-portal-cms/
       context: https://github.com/USSF-ORBIT/ussf-portal-cms.git#main
-      dockerfile: Dockerfile
+      dockerfile: Dockerfile-gh
```
gidjin, line 8: Suggestion: Please change this one back. This is for the cms repo which doesn't have a `Dockerfile-gh` yet. But eventually we will need to change this one at that time too. 

You can see this if you use the `yarn cms:up` or `yarn portal:up` commands, which both use this file. NOTE looks like `docker-compose.client.yml` is missing the following environment var so you'll have to add it to try that command.

`NEXT_PUBLIC_PORTAL_URL=http://localhost:3000`

gidjin (changes requested):
Left one suggestion that needs to be changed before we can merge other wise looks good!

gidjin (approved):
LGTM

docker-compose.client.yml
```diff
@@ -35,6 +35,7 @@ services:
       - LAUNCHDARKLY_SDK_CLIENT_SIDE_ID
       - MATOMO_URL
       - MATOMO_SITE_ID
+      - NEXT_PUBLIC_PORTAL_URL
```
gidjin, line 38: FYI, I added this var since it was missing and preventing the startup

PR #1204, opened 2024-02-01
fix: a11y: edit display name component syntax 

shkeating:
# SC-3150

## Proposed changes

- remove h3 from the Edit Display Name component, it breaks the heading order and the sequence was weird
- `Update name` label should be `update display name`
- `New display name has been saved` instead of `new title has been saved` in success message
- add `required` attribute to update display name, move asterisk to after the label, corrects `id`/`htmlFor` name connection between the `label` and `input`

## Setup

`yarn storybook`
http://localhost:6006/?path=/story/components-editdisplayname--default-edit-display-name

http://localhost:3000/settings

---

## Screenshots

image

image

### DIFF
--- a/src/__tests__/pages/settings.test.tsx
+++ b/src/__tests__/pages/settings.test.tsx
@@ -57,8 +57,6 @@ describe('Settings page', () => {
 
     test('renders the settings page', () => {
       expect(screen.getAllByText('Settings')).toHaveLength(1)
-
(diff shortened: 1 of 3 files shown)


AnnaGingle:
Taking a look now!

AnnaGingle (approved):
Everything called out in the PR works as expected! Great work

gidjin (approved):

PR #1225, opened 2024-03-06
chore: Change test to test:unit for the Gitlab migration

CoryAtTruss:
Change "test" -> "test:unit" for the Gitlab migration.

Not sure how this will interact with existing pipelines, so this PR is only for testing and should not be merged without scrutiny.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "prepare": "husky install",
     "pre-commit": "lint-staged",
     "test:unit": "jest --runInBand --testTimeout=60000",
+    "test": "jest --runInBand --testTimeout=60000",
     "test:watch": "jest --watch  --runInBand --testTimeout=60000",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "lint": "tsc --noEmit && eslint .",
     "prepare": "husky install",
     "pre-commit": "lint-staged",
-    "test": "jest --runInBand --testTimeout=60000",
+    "test:unit": "jest --runInBand --testTimeout=60000",
```
gidjin, line 29: Minor nit: I have a slight preference to keeping the old one in addition for a short time, but we're a small team so adjusting to the new name probably won't take long either. I'll leave it to you.

```suggestion
    "test": "jest --runInBand --testTimeout=60000",
    "test:unit": "jest --runInBand --testTimeout=60000",
```

gidjin (approved):

